PR #5, opened 2025-05-21
Use steamlocate to find path of installed games automatically that have an appid

Tau5:
Steamlocate is a crate that can get the path of the steam data, libraries and steam apps.

We can reduce the times that the user sees the "Select installation path for game" using it when a handler has a steam appid. If we cannot find it we show the popup anyway, but this will allow the player to simply press start without having to search through their filesystem

Steamlocate could also be used in the future to find the steam data path instead of the current method, but that is out of the scope of this PR


wunnr:
Sorry, been busy this weekend. I like the idea of skipping the dialog for Steam games, will take a look at this during the weekend and see if I can merge :+1:

wunnr:
Will make some changes from here but this is a good place to start

PR #19, opened 2025-06-22
Allow single quotes in paths

blckink:
### Summary

- allow single quotes in sanitize_path since they are quoted when launching commands. 
- Otherwise we can't launch games where the executable's file name contains a single quote (for example "Len's Island.exe")


wunnr:
Oops

PR #50, opened 2025-08-17
Add a option to allow multiple instances on the same device

Rudicito:
Hello, I add this option because I think it's useful for testing or for fun?
I'm working on a branch to add the option to use Sway instead of KWin and I wanted to try to play with 64 instances so that's why?

For the PR, I had to fix some issues like: deleting the latest instance of the controller instead of the first one when deleting with button device, or the deleting clickable button deleting the device of the wrong instance, or prevent having the same devices in the same instance.

I Hope it looks not too bad. It's the first time I'm coding in rust.

Also when there is too many instances it goes out of the window, so there should be scrollable thing I guess? Didn't do it in the PR.

### DIFF
--- a/src/app/app.rs
+++ b/src/app/app.rs
@@ -237,14 +237,19 @@ impl PartyApp {
                     {
                         continue;
                     }
-                    if self.is_device_in_any_instance(i) {
+                    if !self.options.allow_multiple_instances_on_same_device 
+                        && self.is_device_in_any_instance(i) {
                         continue;
                     }
 
                     match self.instance_add_dev {
                         Some(inst) => {
-                            self.instance_add_dev = None;
-                            self.instances[inst].devices.push(i);
+                            // Add the device in the instance only if it's not already there
+                            if !self.is_device_in_instance(inst, i) {
+                                self.instance_add_dev = None;
+                                self.instances[inst].devices.push(i);
+                            }
+                            else { continue; }
                         }
                         None => {
                             self.instances.push(Instance {
@@ -284,11 +289,18 @@ impl PartyApp {
         }
     }
 
-    fn is_device_in_any_instance(&mut self, dev: usize) -> bool {
+    fn is_device_in_any_instance(&self, dev: usize) -> bool {
         for instance in &self.instances {
             if instance.devices.contains(&dev) {
                 return true;
             }
+        }
+        false
+    }
+
+    fn is_device_in_instance(&self, instance_index: usize, dev: usize) -> bool {
+        if self.instances[instance_index].devices.contains(&dev) {
+            return true;
         }
         false
     }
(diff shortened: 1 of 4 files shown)


davidawesome02-backup:
While wunnr would have to comment on the acceptance of allowing devices to be accessible devices, this lgtm for controllers, but would NOT WORK on keyboards / mice, the gamescope patch currently binds exclusively on the keyboard / mouse pair (to take it from the desktop env) so sharing it directly with multiple gamescope windows would require a more extensive implementation, here on on gamescope. PLEASE NOTE FROM MY TESTING, YOU CANNOT JUST PASS A FD TO GAMESCOPE (/proc/self/fd/XX) after opening an exclusive lock on the item, LIBINPUT is picky about paths for some reason.

Can you add a note, and explicitly block passing keyboards / mice to multiple devices at the moment?

 
Also:
I applaud your efforts to run 64 games, that's mental. 

Not sure if you want it, but because I am rewriting partydeck (just to show the features I wanted to wunnr, I really hope he adds them to a better implementation), I wrote my own layout manager for `river`, and its the easiest I could find to do layouts like this in.

If you want, I can share that with you as my layout for games, I am trying to patch a few more bugs before I release the code for the small c++ recreation I did (more keyboard / mouse focused, and automated for steam games, while currently partydeck is targeting controllers more, and custom configs.)


I have my layout like this in river, if you want that:

image
image
(as you can tell, I am going into art as a carer path :+1: )

ALSO: please note, if you want my layout, I just pushed the required changes for gamescope to resize to a compositor sizing in my https://github.com/davidawesome02-backup/gamescope/tree/prerelease_david_changes branch (wunnr is currently using the master branch of my repo as the gamescope version, that only has the keyboard patches)

Rudicito:
I push the commit to block multiple keyboard/mice.

Also, river seems a much better choice than sway for this case, didn't know it existed! I'd also want to see your layout please. You can see my partydeck sway branch if you want https://github.com/Rudicito/partydeck/tree/sway!

For the grid, I'm using this formula to get rows and cols number: https://github.com/Rudicito/partydeck/blob/2168115091018add3b23c1fea960d9b6269fccbe/src/grid.rs#L28-L37

davidawesome02-backup:
@Rudicito 
> I push the commit to block multiple keyboard/mice.
> 
> Also, river seems a much better choice than sway for this case, didn't know it existed! I'd also want to see your layout please. You can see my partydeck sway branch if you want https://github.com/Rudicito/partydeck/tree/sway!
> 
> For the grid, I'm using this formula to get rows and cols number: https://github.com/Rudicito/partydeck/blob/2168115091018add3b23c1fea960d9b6269fccbe/src/grid.rs#L28-L37

Here is the current code for my layout if you want to check it out. Please rember, you will have to apply a patch to partydeck with the new layout, or use my gamescope patch to allow dynamic resizing of games (https://github.com/davidawesome02-backup/gamescope/tree/prerelease_david_changes) : https://gist.github.com/davidawesome02-backup/7b535706510690760172f1f08948ddeb


ALSO: PLEASE LET ME KNOW if you have any better ideas for layouts, I like this one, but mby its bad and I am just clouded in my judgement.

(PLEASE TEST THIS BY LAUNCHING PARTYDECK INSIDE RIVER: ```river -c "riverctl spawn river-game-layout; riverctl output-layout river-game-layout; partydeck"```, with my latest version of gamescope, and adding the launch args for gamescope of `--nested-folow-window-scale 1 --force-windows-fullscreen` to properly resize to events fired from riverctl.

wunnr:
Sorry, been super busy irl, will try to test this and get it merged sometime this week.

I also took a look at your sway fork and would love to see that also get merged into main partydeck at some point, and @davidawesome02-backup I'm excited to see how your own c++ implementation turns out; in an ideal world partydeck has support for all the major DEs/window managers

wunnr:
Everything seems to work as expected.

Issue #33, opened 2025-07-15
[Bug] Will not launch on Fedora and Bazitte


Persephone-the-Sheep:
**Describe the bug**
Can't find libbz2.so.1.0

**To Reproduce**
1.  Launch partydeck-rs on Fedora or Bazzite

**Expected behavior**
Program should launch

**Terminal output**
./partydeck-rs: error while loading shared libraries: libbz2.so.1.0: cannot open shared object file: No such file or directory

**Setup Information**
 - Linux Distro: Fedora KDE and Bazzite
 - Hardware: Steam Deck LCD(Bazzite) Intel 12400 Vega 56 (Fedora KDE)

**Additional context**
If i symlink /usr/lib64/libbz2.so.1 to /usr/lib64/libbz2.so.1.0 the program will launch i did this on fedora. Can't do this on bazzite since the file system is read only and i don't know where else to put the symlink. I havent tested any other functionality yet

wunnr:
Seems like this is a common issue with Fedora and Fedora-based distros. 

https://forums.fedoraforum.org/showthread.php?258694-missing-shared-object-file
https://forum.simutrans.com/index.php?topic=21324.0

May be possible to link the program to libbz2.so.1 instead? Probably some compilation option, will look into it

wunnr:
@blckink I noticed your fork had a commit 50592d1. Does that have something to do with the error shown here, or was that to fix a different problem?

wunnr:
@Persephone-the-Sheep can you check if running partydeck with `LD_PRELOAD=/usr/lib64/libbz2.so.1` works? Obviously not ideal, but might be a workaround while I figure out a better solution

Persephone-the-Sheep:
i get the same error when doing `LD_PRELOAD=/usr/lib64/libbz2.so.1 ./partydeck-rs`

zany130:
you can use distrobox too laucnh party deck like this

```
 distrobox-enter -n fedora -- bash -c Applications/PartyDeck-0.5.2/partydeck-rs
```
i tried also to edit the script to use that and it works if you install konsle with `rpm-ostree install konsole`
```
kwin_wayland --xwayland --width $width --height $height --exit-with-session "konsole -e distrobox-enter -n fedora -- bash -c /var/home/zany130/Applications/PartyDeck-0.5.2/partydeck-rs --fullscreen"
```

[...]

PR #51, opened 2025-08-19
Add CI

Twig6943:
title.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,14 +12,14 @@ jobs:
       - name: Check out source code
         uses: actions/checkout@v4
         with:
-          submodules: 'recursive'    # ensures `git submodule update --init` runs
+          submodules: recursive
 
       - name: Install dependencies
         run: |
           sudo apt-get update
           sudo apt-get install -y \
-            build-essential meson cmake make gcc \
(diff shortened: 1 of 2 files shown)


Twig6943:
@wunnr could you also check your email?

davidawesome02-backup:
@Twig6943 

I just have to ask... can you flatten all these patches, and just push a single one, rather than having 27 commits on two files. Also, please provide a real title to this pull request, and commit message on each commit.

I know I don't own this repo, and am not really a contributor, but I would really suggest trying to follow most of the best practices for using git, dont make a new commit for each small change, especially when not merging them instantly, and not changing the message.

wunnr:
Could you change it so that it's built using a Fedora container rather than Arch? I build releases using Fedora right now because of an odd issue where building on Arch causes Fedora users to not be able to run without doing some symlinking of static libraries, whereas binaries built on Fedora seem to work across all distros. See: #33 

Will check my email and respond shortly.

Twig6943:
> Could you change it so that it's built using a Fedora container rather than Arch?

Working on it. But you might want to look into this compiler error (this also happened with arch container)

image

Twig6943:
@wunnr its done.

Twig6943:
ayo?

Issue #58, opened 2025-08-29
[Handler request] Running into issues while creating Celeste handler


IceYetiWins:
I'm working on a Celeste handler that uses the celestenet multiplayer mod and a local server, but there's a few things I'm not sure are possible to do.

**1. Running the local server executable**
I can't seem to find a way to run a separate linux executable along with the game executable. I don't know if that's  currently possible, but that would probably be better than making people set up the server themselves.

**2. Same file being accessed by both instances, causing game to crash**
The game's root folder has a log.txt file that gets deleted temporarily around startup, which means if the other instance tries to access that file at the same time it crashes. Maybe there's a way to create a copy of the file for both instances to access but if there is I couldn't find it. It happens infrequently but requires closing the game and trying again if it happens.

**3. Game save folder can't be copied to profiles**
This is optional as Celeste has multiple save slots. The game's saves are in ~/.local/share/Celeste, which as far as I can tell can't be accessed by partydeck. I tried `"profiles.unique_localshare": true` but that seemed to link the entire localshare folder rather than just ~/.local/share/Celeste.

This is what I have for the handler.json so far:
[handler.json](https://github.com/user-attachments/files/22036436/handler.json)

Soupa7:
So i don't really know about this mod, sorry, but looking at the CelesteNet server github, i don't know if the compatibility prefixes used by party deck have the .Net 7.0, in the past when using non-steam games i've had to install some dependencies using proton tricks. Aside from that, have you tried adding `"game.dll_overrides": ["CelesteNet.server.dll"],` to your handler? Another option may be to install Everest mod loader, set everything up there, and then copying the files into your handler. This is as far as my knowledge goes, sorry.

<img width="955" height="321" alt="Image" src="https://github.com/user-attachments/assets/a59c36ae-9915-44fd-82bd-8388c61e1afd" />

IceYetiWins:
I just built the celestenet server executable myself and was planning to include it in the handler. Is there a way to run a standard linux executable through a partydeck handler? For the client side mod the user would have to install it, as the mod loader completely rewrites a lot of the game files.

wunnr:
The launcher currently doesn't have the ability to run a single server executable separate from the game instances, I'll look to implement that at some point

>The game's saves are in ~/.local/share/Celeste, which as far as I can tell can't be accessed by partydeck. I tried "profiles.unique_localshare": true but that seemed to link the entire localshare folder rather than just ~/.local/share/Celeste.

[...]

PR #61, opened 2025-09-03
Add handler option for custom pause between instance starts

Rudicito:
https://github.com/user-attachments/assets/a915d89c-4179-4a3a-9ff8-3648cdc502c8

This add the option `game.pause_between_starts` in the handlers json.
Example: `"game.pause_between_starts": 5`

It could be used for example in Celeste handler (#58).

Hope the PR looks okay!

### DIFF
--- a/src/handler.rs
+++ b/src/handler.rs
@@ -28,6 +28,7 @@ pub struct Handler {
     pub copy_instead_paths: Vec<String>,
     pub remove_paths: Vec<String>,
     pub dll_overrides: Vec<String>,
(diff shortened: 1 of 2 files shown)


wunnr:
Sorry for the late response, Silksong has been terrible for my productivity.
Everything looks fine so I'll merge but I'll let you know that after doing multi-monitor support I plan to do an overhaul on how games are launched, so this option will definitely become deprecated at some point.

Rudicito:
No problem, don't worry

PR #65, opened 2025-09-09
Added more cli functionality

Blahkaey:
I added the following cli args, these allow the user to specify the handler, profiles, and devices so the launch process is less repetitive for every game launch.

--handler        
Launch a game using its handler uid

--player          
Add a player with profile and devices
Format: profile=name,devices=id,id
Optional: monitor=index
Profile is auto created if it doesn't exist
Devices take either the path or name of the device

--auto-launch            
Automatically start the game without gui interaction (still shows the app_light screen)

Here is an example of what i used for one of my tests:

./partydeck --handler "t6zmpluto" \
--player "profile=Player1,devices=Xbox Controller" \
--player "profile=Player2,devices=Logitech G502,CORSAIR CORSAIR K63 Wireless USB Receiver Keyboard"

I tested it with 2 players a few handlers in desktop and game mode and it worked well,  I have not tested specifying a monitor.

Also thanks for the work so far, the project is already awesome. i have recently moved to linux and there was no proper split screen app like nucleus coop. i am already enjoying this app alot more than nucleus, especially now that i can automate the launching of a split screen session


wunnr:
Thanks for the kind words, and I appreciate the work you've put in here. Unfortunately I'm currently in the process of reworking the entire game launching code and it's looking like that work will probably break a lot of the stuff you've done here, so I won't be merging this as of now. Once the new game launching systems are done and merged into main feel free to revisit this and see if any of the features here are still necessary

PR #121, opened 2025-12-24
[Draft] Add AppImage

Twig6943:
AppImage has these 2 errors;

image

image

(PR name has flatpak because I might start on a flatpak depending on how stuff goes)


Twig6943:
Do you know how to resolve these issues? @wunnr

davidawesome02-backup:
As far as I am aware, flatpak will not work for a few reasons, mainly because of flatpak's permission model, flatpak is not able to read / write to other's locations

Also due to use of bwrap in this project, we would have to confirm nested kernel namespaces work properly with flatpak. This appears to be stalled possibly waiting on the kernel "landlock" feature, but this is new as of this year, and I am not sure if I am misreading the github and kernel mailing list.

```
Ideally, Flatpak would simply support nested namespacing and nested sandboxes, but currently it does not. 
Flatpak uses seccomp to prevent applications in a sandbox from having direct access to user namespaces. 
There is an API that can be used to create a sub-sandbox, but it is more restrictive. 
He said that the restrictions to user namespaces are outdated: 
""for a long time it wasn't really a good idea to expose user namespacing because it exposed a big kernel API to user space that could be exploited"". 
Wick feels that user namespaces are, nowadays, a well-tested and a much-used interface. 
He does not think that there is much of a good argument against user namespaces anymore.
``` 
\- https://lwn.net/Articles/1020571/ (2025)
https://github.com/flatpak/flatpak/issues/5921?timeline_page=1


flatpak SHOULD be able to run commands on the host as an alternative with `flatpak-spawn`(https://docs.flatpak.org/en/latest/flatpak-command-reference.html#flatpak-spawn) if needed to run main steam games, but running flatpak steam this way is verrrry round about, and most likely wont get us into the stores due to us esencialy breaking the container Note: [specified dbus requirements](https://docs.flatpak.org/en/latest/sandbox-permissions.html#d-bus-access) . I would possibly suggest a small loader in the form of a [extention](https://docs.flatpak.org/en/latest/dependencies.html#extensions) as documented [here](https://docs.flatpak.org/en/latest/extension.html) to inject into the steam flatpak, and expose a dbus api for starting the steam games there, but that requires nested sandboxes and....

This is a lot of work, I don't know if it is fully worth it. Steam decks should be able to use the stand alone executable (I currently have a patch that allows embedding all the required files into a single executable that is stalled waiting on my other patch and wunnr's approval)

While I see the point of a flatpak especially for steamdeck, I think this is a large problem for development here, and I think appimage could be just the standalone executable unless you have another reason to use appimage format.

Sorry to be a downer about this, I really like the idea of properly shipping this, and it is something I was thinking about recently too, happy to get any more ideas though! Maybe I am incorrect about flatpak, that is the ideal launch location due to kde's "discover", but it may not be possible.

lastly, as for the errors above, you didnt bundle `splitscreen_kwin.js` nor `splitscreen_kwin_vertical.js` as is required for the kde split screen enabled by default in settings.

Hopefully we can get wunnr's opinion!

Twig6943:
Soo tldr; flatpak seems to be a no no

> and I think appimage could be just the standalone executable unless you have another reason to use appimage format.

AppImage is to ship it properly without relying on distro specific packages.

> lastly, as for the errors above, you didnt bundle `splitscreen_kwin.js` nor `splitscreen_kwin_vertical.js` as is required for the kde split screen enabled by default in settings.

Thanks for the info. Will see about that.

davidawesome02-backup:
> > and I think appimage could be just the standalone executable unless you have another reason to use appimage format.
> 
> AppImage is to ship it properly without relying on distro specific packages.

just raw binaries work on all linux systems, even more compatible than appimage hence my recommendation.

Okay, if you think appimage is a good format, I am still on the side of targeting direct binaries as there is no real reason to have the appimage overhead except maybe added compression that is hard for me to add to the embedded binaries easily... but I suggest we wait for wunnr's response.

Twig6943:
> just raw binaries work on all linux systems, even more compatible than appimage hence my recommendation.

any package manager/s that you know of that supports binaries? (Also technically wrong those wont run on the best distro known to man nixos /j )

davidawesome02-backup:
> any package manager/s that you know of that supports binaries? (Also technically wrong those wont run on the best distro known to man nixos /j )

I mean... who said we were gonna use a package manager in my solution? I would recommend just shipping the raw executable binary, and bypassing any package managers as a whole. If we were to use a single binary, it would be able to work nixos too lol, and most distro package managers are just wrappers for zip with binaries in them, here it would just be a single binary rather than multiple.

I get we are joking at this point, but we should figure out how exactly we want to take on the packaging in a reasonable way. I like a single binary because I have enjoyed what yt-dlp dose for their distribution pattern on linux, but I am not sure it fits how wunnr wants his due to partydeck being a GUI based one. Main thing is "discover" compatibility, but I am not sure if we can make ours an appimage without the run as host hacks... but those may work.

davidawesome02-backup:
```
Submissions incompatible or restricted by Flatpak sandbox
Certain types of submissions, such as development tools and IDEs with a broad
scope are generally not well-suited for Flatpak due to inherent sandboxing limitations and 
severely degrade the user experience. Unless these applications are being officially 
submitted by the upstream developers and/or upstream has explicitly committed to 
supporting Flatpak they will not be accepted.
```

```
Unacceptable software design practices
Submissions that include antipatterns or insecure design choices, for example, 
disabling or ignoring security checks, insecure cryptography practices, exposing 
sensitive information, overly permissive configurations etc. will not be accepted.
```
\- https://docs.flathub.org/docs/for-app-authors/requirements

BUT... they would possibly accept it, because its not explicitly stated, and we follow all other requirements... so this could work as a way to get it to the people using steam decks easily.

for wunnr if he wants to try to do this: https://docs.flathub.org/docs/for-app-authors/submission

but just know it will be kinda hard, so maybe using a binary or if you prefer appimage is still easiest.

Samueru-sama:
> just raw binaries work on all linux systems, even more compatible than appimage hence my recommendation.

That's only true if the binary is fully static. 

If not static then the best next solution is [sharun](https://github.com/VHSgunzo/sharun) with the [uruntime](https://github.com/VHSgunzo/uruntime)

I have a few demos apps here using that [here](https://github.com/pkgforge-dev/Anylinux-AppImages/releases/tag/demo) the only requirements are a shell in `/bin/sh` and write access to `/tmp`, they are independend of the host libc, you can test them in several containers to see.

davidawesome02-backup:
> > just raw binaries work on all linux systems, even more compatible than appimage hence my recommendation.
> 
> That's only true if the binary is fully static.
> 
> If not static then the best next solution is [sharun](https://github.com/VHSgunzo/sharun) with the [uruntime](https://github.com/VHSgunzo/uruntime)
> 
> I have a few demos apps here using that [here](https://github.com/pkgforge-dev/Anylinux-AppImages/releases/tag/demo) the only requirements are a shell in `/bin/sh` and write access to `/tmp`, they are independend of the host libc, you can test them in several containers to see.

True, let me check my current build's runtime dependencies, all are available on a base install of arch linux with gamescope installed if I rember correctly. I ment that in the way of its able to work on all systems, and the dependancy requirements are minimal outside of normal deps. We do have to remove some of the current ones still though. I have a few more to remove, but a lot are removed in my super large pull request.

As for partydeck, its only requirements I can find are:
```
        linux-vdso.so.1 (0x00007f9c6f92d000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f9c6f8d9000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f9c6e310000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f9c6e11a000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f9c6f92f000)
``` 
(totally could be linked static, but I think it may be missing sdl2 for some reason....)
`std::process::Command::new("zip")` /src/util.rs#L95-L102 (could be the zip crate, or if wunnr wouldn't mind breaking back compat, we could switch from zip and go to only lzma or only zstd to get even smaller without requiring 10 (exaggeration, but you know what I mean) different compression libs as zip dose.
`Command::new("mountpoint").arg(dir).status()` /src/util.rs#L129
`Command::new("umount")` /src/util.rs#L151
`.map(|_| Command::new("fuse-overlayfs"))` /src/launch.rs#L447

And then just the requirements for gamescope. 

River (my addition not merged yet) requires nothing really and would be optional, so that is not hard. Umu / GBE_FORK are the ones I dont know about, but are all optional.

wunnr:
Not familiar with how packaging with appimage/flatpak works but if you can make a working proof of concept I'll certainly consider it. As Davidawesome said I don't think flatpak would work without full disk access permissions which sort of defeats the purpose of having a flatpak, though I guess having it on the discover store would be nice.

Btw @Twig6943 if you are just building through the build.sh script your build is likely missing stuff like Goldberg because the script is sort of incomplete; will try to get a better one out soon but in the meantime you can use the latest release as reference for the folder structure.

Twig6943:
> Not familiar with how packaging with appimage/flatpak works but if you can make a working proof of concept I'll certainly consider it. As Davidawesome said I don't think flatpak would work without full disk access permissions which sort of defeats the purpose of having a flatpak, though I guess having it on the discover store would be nice.
> 
> Btw @Twig6943 if you are just building through the build.sh script your build is likely missing stuff like Goldberg because the script is sort of incomplete; will try to get a better one out soon but in the meantime you can use the latest release as reference for the folder structure.

bro thats literally what I did because of that lol literally check appimage.sh

Twig6943:
> lastly, as for the errors above, you didnt bundle `splitscreen_kwin.js` nor `splitscreen_kwin_vertical.js` as is required for the kde split screen enabled by default in settings.

Got it to bundle properly

https://github.com/Twig6943/partydeck/commit/46358982830f17041ed9a531d63a40376d80e911

But now I gotta solve this error which I have no clue what's causing

image

davidawesome02-backup:
> > lastly, as for the errors above, you didnt bundle `splitscreen_kwin.js` nor `splitscreen_kwin_vertical.js` as is required for the kde split screen enabled by default in settings.
> 
> Got it to bundle properly
> 
> [Twig6943@4635898](https://github.com/Twig6943/partydeck/commit/46358982830f17041ed9a531d63a40376d80e911)
> 
> But now I gotta solve this error which I have no clue what's causing


Just so you know; its due to us (sorry, not to be mean, but we do need to fix this) currently catching an error WAY too high, and not logging all the information.

Its caused by the lines: https://github.com/wunnr/partydeck/blob/62a0973646ecc0f6750531345af90e62d372ba7d/src/app/app.rs#L448C1-L451C18
```rust
                if let Err(err) = launch_game(&handler, &dev_infos, &instances, &cfg) {
                    println!("[partydeck] Error launching instances: {}", err);
                    msg("Launch Error", &format!("{err}"));
                }
```
just taking all errors and printing them instantly. I am looking for resolutions, but I am not sure if wunnr would want a pr for something this small, I may just look and make a .patch, or let him fix it.

Now, of the things that can cause that error for your part @Twig6943 :
that's caused by 

```rust
    for mut cmd in new_cmds { // we should rewrite this as a indexed loop too lol
        let handle = cmd.spawn()?; // <-- this guy right here
        // ...
    }
```

And that tells me something is not in your path, either `“/tmp/.mount_partyremp11889608271946847563/bin/bin/gamescope-kbm"`(incorrect because I used ocr, I am not retyping this) or `bwrap` or `“/shome/twig/.local/share/partydeck/tmp/game-8/sm64coopdx"` (again ocr) (this one confuses me, because the game shouldnt be inside your partydeck tmp directory...)

Please check what is accessible in your path, or make sure you have bwrap properly installed, or check the handler.


EDIT: looking a little more shows that the `/bin/bin` may be caused by you putting `partydeck` at the same level as `gamescope-kbm`, please put `gamescope-kbm` in a folder called `bin` right next to `partydeck` (the folder is next to partydeck, not `gamescope-kbm` I think, or change the code on https://github.com/wunnr/partydeck/blob/62a0973646ecc0f6750531345af90e62d372ba7d/src/paths.rs#L52:
```rust
pub static BIN_GSC_KBM: LazyLock<PathBuf> = LazyLock::new(|| {
    let bin_candidates = [PathBuf::from("/usr/bin"), PathBuf::from("/usr/local/bin")];

    for candidate in &bin_candidates {
        let bin = candidate.join("gamescope-kbm");
        if bin.exists() {
            return bin;
        }
    }

    let bin = env::current_exe().unwrap().parent().unwrap().join("bin");
    bin.join("gamescope-kbm")
});
``` 
^^^^^ wunnr, I changed this to use actual path search in my embedded binaries one, so if you want we can use that. Mostly what you want to do is prepend the `./bin` and then walk all path vars to find it.
)

Twig6943:
How does this commit look? https://github.com/Twig6943/partydeck/commit/e6820021703a8af8b31b3de451b687dc7cce12c1

pkg/appimage/AppImage.sh
```diff
@@ -25,9 +25,12 @@ wget --retry-connrefused --tries=30 "$SHARUN" -O ./quick-sharun
 chmod +x ./quick-sharun
 
 # Point to binaries
-./quick-sharun ./partydeck ./bin/gamescope-kbm ./bin/umu-run /usr/bin/fuse-overlayfs
+./quick-sharun ./partydeck ./bin/gamescope-kbm ./bin/umu-run /usr/bin/fuse-overlayfs /usr/bin/bwrap
 
 # Copy rest
+mkdir -p ./AppDir/bin/bin
+cp ./bin/gamescope-kbm ./Appdir/bin/bin
```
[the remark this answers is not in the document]
Twig6943: > TLDR: `mkdir -p ./AppDir/bin/bin && ln -f ./AppDir/sharun ./AppDir/bin/bin/gamescope-kbm`

Added; https://github.com/Twig6943/partydeck/commit/b29480899d296f5cba83bfdfa6b1c70a237c14e8

pkg/appimage/AppImage.sh
```diff
@@ -0,0 +1,46 @@
...
+SHARUN="https://raw.githubusercontent.com/pkgforge-dev/Anylinux-AppImages/refs/heads/main/useful-tools/quick-sharun.sh"
+
+export ADD_HOOKS="self-updater.bg.hook"
+#export UPINFO="gh-releases-zsync|${GITHUB_REPOSITORY%/*}|${GITHUB_REPOSITORY#*/}|latest|*$ARCH.AppImage.zsync"
+export OUTNAME=partydeck-anylinux-"$ARCH".AppImage
+export DESKTOP=partydeck.desktop
+export ICON=./partydeck.png
+export DEPLOY_OPENGL=0
+export DEPLOY_VULKAN=0
+export DEPLOY_DOTNET=0
```
Samueru-sama: these should not be set to `0` unless there is a good reason to disable them.

Twig6943: New update: https://github.com/Twig6943/partydeck/commit/ab5767fea5971b9c46e4c3ce200111ee5e462b1f

Twig6943:
Good news; the app works now;

https://i.imgur.com/OA90SDk.mp4

Twig6943:
Things blocking this pr from being merged (just noting);

app looking for bins straight in a relative folder named bin
more testing

wunnr:
My bad, I already have it search /usr/bin for the gamescope-kbm/umu binaries, I didn't think to just search through PATH. Will fix soon

davidawesome02-backup:
@wunnr 
> My bad, I already have it search /usr/bin for the gamescope-kbm/umu binaries, I didn't think to just search through PATH. Will fix soon

When I was making my embedding, I used the `search_path = "0.1.4"` crate to do it, where its api is used like this:

(This is from my embeded branch, so dont just use this, but if we wanted an option to use embeded / bundled or system executable this is how I think it should be done, and has worked in my tests)

```rust
    if cfg.enable_kwin_script {
        let script;
        
        let mut search_path_obj = search_path::SearchPath::new("PATH")?;
        if cfg.use_embed_executables {
            search_path_obj.append(PATH_PARTY.join("extracted_bin"));
        }

        if instances.len() == 2 && cfg.vertical_two_player {
            script = search_path_obj
                .find_file(Path::new("splitscreen_kwin_vertical.js"))
                .expect("Failed to locate kde splitscreen script in path");
        } else {
            script = search_path_obj
                .find_file(Path::new("splitscreen_kwin.js"))
                .expect("Failed to locate kde splitscreen script in path");
        };

        layout_manager::kwin_dbus_start_script(script)?;
    }
 ```
 
 Also if we are moving things around to search path instead, we may want to change the current name `gamescope-kbm` to just `gamescope` inside our bundle and have a `use embeded / bundled` option as well as `enable gamescope kbm features` so people (like me) can test with gamescope that is in their PATH as well, or if the changes get merged upstream (prob not at this point), just use that.

**Comment by wunnr** [comment]
@Twig6943 fixed in a31f0e6, please test to see if everything works. As for the res files, the app also first searches `/usr/share/partydeck` before `[partydeck]/res`, so if you're building appimage I think you would put goldberg files and the kwin script files in `./AppDir/usr/share/partydeck`? If that doesn't work or that isn't an appropriate location to check for res let me know

**Comment by Samueru-sama** [comment]
> @Twig6943 fixed in [a31f0e6](https://github.com/wunnr/partydeck/commit/a31f0e6e80c4dacd2ee4083d571277773465b150), please test to see if everything works. As for the res files, the app also first searches `/usr/share/partydeck` before `[partydeck]/res`, so if you're building appimage I think you would put goldberg files and the kwin script files in `./AppDir/usr/share/partydeck`? If that doesn't work or that isn't an appropriate location to check for res let me know

`/usr/share/partydeck` is fine, but ideally the app should check [`XDG_DATA_DIRS`](https://specifications.freedesktop.org/basedir/latest/#variables) first to find its data directory and when the var is not set fallback to `/usr/local/share:/usr/share`.

Also the app should first check the `res` directory before `/usr/share/` to avoid conflicts if the user happens to have it installed already.

But in any case this is not a problem since `quick-sharun` will detect and patch any hardcoded `/usr/share/partydeck` path from the binary.

**Comment by Twig6943** [comment]
> @Twig6943 fixed in [a31f0e6](https://github.com/wunnr/partydeck/commit/a31f0e6e80c4dacd2ee4083d571277773465b150), please test to see if everything works. As for the res files, the app also first searches `/usr/share/partydeck` before `[partydeck]/res`, so if you're building appimage I think you would put goldberg files and the kwin script files in `./AppDir/usr/share/partydeck`? If that doesn't work or that isn't an appropriate location to check for res let me know

I need you to make a new release for the binary to have said commit. So let me know when you have (just mark it as pre-release so regular people don't end up downloading)

**Comment by wunnr** [comment]
@Twig6943 prerelease build uploaded, let me know if there are any issues

**Comment by Twig6943** [comment]
Alright will give it a go tomorrow

Update: got busy with [something else](https://github.com/nocss42/GardenGate) will try to get this done this week

**Comment by Twig6943** [comment]
@wunnr 

The appimage prerelease you made for this doesn't have the same structure as the current release 0.8.5

On the left is the current release on the right is what you made which doesn't have the binary in the same location

image

Also what's this whole `GNUSparseFile.0` abomination and the file inside it;

image

image

**Comment by wunnr** [comment]
Strange, that definitely was not what it looked like when I built it. Will reupload a correct version tonight

**Comment by wunnr** [comment]
@Twig6943 based on a cursory search on GNUSparseFile it seems my computer compressed the tar with settings that are not compatible with certain tar uncompress tools? Will have to investigate further but for now I've reuploaded it as a zip on the same prerelease page.

**Comment by Twig6943** [comment]
@wunnr do you know where the scripts that fail do and where are they stored in the release?

image


In other news the app does use gamescope binary from just `/bin` now;

image

**Comment by wunnr** [comment]
@Twig6943 if you're talking about the gamescope script not found messages, I believe those are scripts related to Steam Deck displays, as in my testing they always fail when using gamescope on a PC. This is just a warning gamescope emits and does not actually cause any problems.

**Comment by Twig6943** [comment]
> @Twig6943 if you're talking about the gamescope script not found messages, I believe those are scripts related to Steam Deck displays, as in my testing they always fail when using gamescope on a PC. This is just a warning gamescope emits and does not actually cause any problems.

Well other than that it worked fine on me please give it a shot youself as well

(Clone my repo go to pkg/appimage, extract the latest zip you made for this there run AppImage.sh and it'll generate an AppImage )

**Comment by SkyLeite** [comment]
Hey @Twig6943, would it be possible for you to add the AppImage to the thread so you can get more feedback?

**Comment by wunnr** [comment]
@SkyLeite I just uploaded the appimage to the prerelease page [here](https://github.com/wunnr/partydeck/releases/tag/appimage-testing).

**Comment by Twig6943** [comment]
Soo any noticeable issues?

**Comment by wunnr** [comment]
Just tested, does not work on my system. 

```
[gamescope-kbm] [Info]  wlserver: [xwayland/server.c:107] Starting Xwayland on :2
MESA-LOADER: failed to open dri: libgallium-25.3.3-arch1.1.so: cannot open shared object file: No such file or directory (search paths /tmp/.mount_partyremp4474297226716082243/shared/lib/gbm, suffix _gbm)
couldn't create gbm device
XWAYLAND: Disabling GLAMOR support
EGL setup failed, disabling glamor
Failed to initialize glamor, falling back to sw
MESA-LOADER: failed to open dri: libgallium-25.3.3-arch1.1.so: cannot open shared object file: No such file or directory (search paths /tmp/.mount_partyremp4474297226716082243/shared/lib/gbm, suffix _gbm)
MESA-LOADER: failed to open dri: libgallium-25.3.3-arch1.1.so: cannot open shared object file: No such file or directory (search paths /tmp/.mount_partyremp4474297226716082243/shared/lib/gbm, suffix _gbm)
MESA-LOADER: failed to open dri: libgallium-25.3.3-arch1.1.so: cannot open shared object file: No such file or directory (search paths /tmp/.mount_partyremp4474297226716082243/shared/lib/gbm, suffix _gbm)
[gamescope-kbm] [Info]  vblank: Using timerfd.
```

This happens when gamescope is initializing, and causes any game to either crash on startup or have broken, flickering graphics. Running on Arch, AMD card with mesa, all drivers up to date.

**Comment by Twig6943** [comment]
> Just tested, does not work on my system.
> 
> ```
> [gamescope-kbm] [Info]  wlserver: [xwayland/server.c:107] Starting Xwayland on :2
> MESA-LOADER: failed to open dri: libgallium-25.3.3-arch1.1.so: cannot open shared object file: No such file or directory (search paths /tmp/.mount_partyremp4474297226716082243/shared/lib/gbm, suffix _gbm)
> couldn't create gbm device
> XWAYLAND: Disabling GLAMOR support
> EGL setup failed, disabling glamor
> Failed to initialize glamor, falling back to sw
> MESA-LOADER: failed to open dri: libgallium-25.3.3-arch1.1.so: cannot open shared object file: No such file or directory (search paths /tmp/.mount_partyremp4474297226716082243/shared/lib/gbm, suffix _gbm)
> MESA-LOADER: failed to open dri: libgallium-25.3.3-arch1.1.so: cannot open shared object file: No such file or directory (search paths /tmp/.mount_partyremp4474297226716082243/shared/lib/gbm, suffix _gbm)
> MESA-LOADER: failed to open dri: libgallium-25.3.3-arch1.1.so: cannot open shared object file: No such file or directory (search paths /tmp/.mount_partyremp4474297226716082243/shared/lib/gbm, suffix _gbm)
> [gamescope-kbm] [Info]  vblank: Using timerfd.
> ```
> 
> This happens when gamescope is initializing, and causes any game to either crash on startup or have broken, flickering graphics. Running on Arch, AMD card with mesa, all drivers up to date.

[Added OpenGL](https://github.com/Twig6943/partydeck/commit/13a9d2d16748420b8b51df59ef5f86ef257d73c2) please test newer build (might add vulkan as well depending on how it goes)

**Comment by wunnr** [comment]
`export DEPLOY_OPENGL=1` did fix OpenGL games. Vulkan is still broken even after adding `export DEPLOY_VULKAN=1`. When running proton/dxvk games I get `pressure-vessel-wrap[317479]: W: Failed to load Vulkan ICD #0 from /tmp/.mount_partyremp12439648318735807112/share/vulkan/icd.d: openat(/tmp/.mount_partyremp12439648318735807112/share/vulkan/icd.d): No such file or directory`, when running native vulkan games they silently crash with no relevant terminal output.

**Comment by Twig6943** [comment]
@wunnr Added [some stuff](https://github.com/Twig6943/partydeck/commit/6a14cae2fc5abb8ca3328ac5db735eca0b8bf595) for vulkan, give it a shot

**Comment by wunnr** [comment]
Doesn't seem to fix it for me. Interestingly, I checked for the icd file that fails to open (`/tmp/.mount_partyremp[..]/share/vulkan/icd.d`) because of `no such file or directory`, and it does actually exist; I don't know why pressure-vessel-wrap from Proton (and maybe native games) cannot open it.

**Comment by Samueru-sama** [comment]
Most likely `bwrap` is not binding the path to the icd files.

We had a similar issue with glycin.

**Comment by wunnr** [comment]
Quick update: found out that it was because of the usage of `--tmpfs /tmp` args when launching the games with bwrap. Some native linux games put lock files in /tmp, so we give each game instance its own /tmp so they don't see each other. However, since the icd files live at `/tmp/.mount_partyremp[..]`, creating the tmpfs means that the game instances weren't able to locate the relevant files. I threw together a quick fix that simply looks for the existence of  `/tmp/.mount_partyremp[..]` and, if it does exist, binds it into each instance's tmpfs.

However, this fix did not allow games to run. The new error I get is: `pressure-vessel-wrap[330960]: W: Failed to load Vulkan ICD #0 from /tmp/.mount_partyremp[..]/share/vulkan/icd.d: Is a directory
`. The icd directory is recognized, but the application is trying to read the icd.d folder as an icd instead of one of the json files *inside* icd.d. I have no clue why this is; maybe something to do with hardcoded paths getting patched in the appimage creation process? If someone has an example of a vulkan-based app/game being properly deployed as an appimage I might be able to better understand what's going on here

**Comment by Samueru-sama** [comment]
What happens if you change `TMPDIR` to a different location? the appimage will mount there instead. 

> The icd directory is recognized, but the application is trying to read the icd.d folder as an icd instead of one of the json files inside icd.d.

That is werid, application should not be doing icd discovery themselves. [This is handled by vulkan itself](https://github.com/KhronosGroup/Vulkan-Loader/blob/main/docs/LoaderDriverInterface.md#driver-discovery-on-linux)

> I have no clue why this is; maybe something to do with hardcoded paths getting patched in the appimage creation process?

I don't think so, we don't patch `/usr/share/vulkan` since it doesn't need to  be patched, the vulkan loader correctly checks `XDG_DATA_DIRS` and a ton of other variables. 


> If someone has an example of a vulkan-based app/game being properly deployed as an appimage I might be able to better understand what's going on here

* https://github.com/hrydgard/ppsspp/releases/tag/v1.19.3
* https://github.com/eden-emulator/Releases/releases
* A ton of games listed [here](https://github.com/pkgforge-dev/Anylinux-AppImages)
* [and more that haven't been added](https://github.com/orgs/pkgforge-dev/repositories?q=Link4Electronics)

**Comment by Samueru-sama** [comment]
> I threw together a quick fix that simply looks for the existence of `/tmp/.mount_partyremp[..]` and, if it does exist, binds it into each instance's tmpfs.

you can also check for `APPDIR` env variable which contains the mountpoint path.

**Comment by Twig6943** [comment]
bump? (been a while)

**Comment by Samueru-sama** [comment]
I don't think the `pressure-vessel-wrap` issues are fixable btw. 

I recently tried to package [Lutris using sharun](https://github.com/Samueru-sama/Lutris-sharun-test) and ran into the very same issue. 

In theory setting `PRESSURE_VESSEL_FILESYSTEMS_RO=${SHARUN_DIR}` would fix these issues but it doesn't, it still complains that the several locations inside the appimage are unlikely to appear in `/run/host`

**Comment by davidawesome02-backup** [comment]
@Samueru-sama 

I can see if its possible for me to take a look into this in a few days. Is the issue you see with `pressure-vessel-wrap` is that still only the `pressure-vessel-wrap[330960]: W: Failed to load Vulkan ICD #0 from /tmp/.mount_partyremp[..]/share/vulkan/icd.d: Is a directory` issue? If so, we should just have to expose that to the system as a full absolute path, or as a symlink to the real icd. That must come from the main system, so we have to provide that /usr/share/vulkan/icd.d is the real directory for vulkan to work properly with nearly any application as far as I can tell at a brief look, and is dependent on the host system gpu.

I think we can accept that loss in portability, so any system that has vulkan installed can run the appimage still, if that's acceptable you all and the core issue here still.

EDIT: if that is the issue, we need to take a look at 

```    SHARUN_ALLOW_SYS_VKICD=1       Enables breaking system vulkan/icd.d for vulkan loader``` -- https://github.com/VHSgunzo/sharun or the ```VK_DRIVER_FILES 	${SHARUN_DIR}/share/vulkan/icd.d``` env var they set

Sorry I didn't bump this earlier, got distracted with the other sub-projects.

**Comment by wunnr** [comment]
IMO we should just use the system vulkan ICD; partydeck is only targeting systems that already have vulkan anyways so I don't see any benefit to having it bundled

**Comment by Samueru-sama** [comment]
Alright, be aware that using the host vulkan means that the appimage will only work on glibc systems.

And it also means weird bugs [like this](https://codeberg.org/pkgforge-dev/Citron-AppImage/issues/14) which are caused by older versions of vulkan in the host.

I'm also not sure if this fixes the issue, `SHARUN_ALLOW_SYS_VKICD` only takes effect if the appimage has a vulkan icd in it, and it wasn't the case here originally iirc.

**Comment by davidawesome02-backup** [comment]
Okay, finaly got it to run... ish on my system, but we are missing some dependencies. I think I will make this into a docker build image or just a github compile script, so we don't have to have different environments we are testing with. Currently, `gamescopereaper` isn't included in the files we require, and the program on my computer didn't run unless (somehow) I enabled debug symbols to try to identify why I was getting a segfault inside the rust binaries.

So far I have these packages being the ones required on top of the base arch linux install to build, but I may be missing some:
```
sudo pacman -S --noconfirm git meson rust cargo clang glslang libcap wlroots0.18 sdl2 vulkan-headers libx11 libxmu libxcomposite libxrender libxres libxtst libxkbcommon libdrm libinput wayland-protocols benchmark xorg-xwayland pipewire cmake libavif libheif aom rav1e libdecor libxdamage luajit libva libxss wget curl xorg-server-xvfb fuse-overlayfs bubblewrap libxkbcommon zsync pipewire pulseaudio pipewire-audio alsa-plugins libxkbcommon-x11 libvdpau libvdpau-va-gl
```
To get a clear system, I am currently testing on github codespaces, but I will be moving to docker in a few.

davidawesome02-backup:
Okay, got it to work and compile in a docker container / or in a github action. It works currently but still the vulkan ICD is not working for programs inside of the created container, it seams to be something about us including vulkan at all that breaks the program... gamescope requires vulkan, and it looks to be patched properly to run with the required tweaks inside this strange environment, but the real programs / games (even vulkaninfo) fail due to the incorrect path or environment not loading vulkan entirely. I will see if I can just have vulkan entirely not bundled so system vulkan is just required and it works properly via that. Not sure yet, but will keep you guys updated. 

If anyone wants to compile right now, the minimally required patches and everything to get around the recent Arch linux break (wlroots relied on a enum that changed in upstream libinput), checkout my [partydeck fork](https://github.com/davidawesome02-backup/partydeck/tree/flatpak-tests-2) as of now. (My github workflow testing versions have been done on a private repo as to modifying my `main` branch right now, so if you want to run that, make a fork and set that as the main branch, or just use the docker file included after building and applying the patch by hand)

davidawesome02-backup:
Quick update, found out why vulkan games don't "just work" right now, and that's due to the vulkan overrides that are applied to the container that causes the vulkan icd and other required files to be placed in `/tmp/.mount_partyremap*` (As wunnr said and I skipped over on accident) and of course we use bwrap that creates a new temp directory and thus hides the partydeck remapped files for vulkan. I am going to try compiling without vulkan at all enabled, so the base system one is used so we don't have to have special handling, but if that doesn't work  best here, I can add the code to pass in the /tmp files, or remap somewhere else if that can work somehow...

EDIT 1:
You can use the current build and just unset `VK_DRIVER_FILES` and any vulkan program will work right now. Maybe this is ideal as it uses our vulakn env for our project, and then if there is a vulkan game it runs with system vulkan just like it normally would.

davidawesome02-backup:
Does anyone want to test my version to see if it works for you guys? I have not yet tested on a full base arch install, but it seems to work on base steamos, and my testing arch repo. If you want I can provide a build or I can publish on my branch a workflow build, or you can build from the docker file provided.

Let me know if you guys don't like how I am doing this to make it work: here are [my changes](https://github.com/partydeck/partydeck/compare/main...davidawesome02-backup:partydeck:appimage-tests)

Samueru-sama:
> You can use the current build and just unset `VK_DRIVER_FILES` and any vulkan program will work right now. Maybe this is ideal as it uses our vulakn env for our project, and then if there is a vulkan game it runs with system vulkan just like it normally would.

Interesting. 

Yeah sharun automatically sets `VK_DRIVER_FILES` to the path of each icd file, maybe this is the reason it breaks since the path in the container is going to be different. 

Btw you don't need to `unset VK_DRIVER_FILES` in the code of the application, you can make a `.env` file in the `AppDir` that contains `unset VK_DRIVER_FILES`

TLDR: `echo 'unset VK_DRIVER_FILES' >> ./AppDir/.env`, the `>>` is very important here since that file can have other vars that are set by `quick-sharun` when needed.

davidawesome02-backup:
The main thing is that this approach runs the main program and `gamescope` using the provided vulkan binary, but the games use the system vulkan binary. This means any openGL games run without a system vulkan, where `gamescope` runs under the bundled vulkan, and any games launched use native unknown vulkan so it actually matches anything on the base system.

If this is somehow an issue let me know, but I think this is the best way due to the program's structure.

(Sorry if this is confusing writing this quickly between some other things I am doing)

Blahkaey:
@davidawesome02-backup i tested your app image branch. the gh action worked fine and the appimage ran on my system without any issues

davidawesome02-backup:
@Blahkaey Cool, I will get it cleaned up and try to make a PR for it in a few days. 

@Twig6943 do you want to test it before then to see if I am missing anything you wanted here? 

If I am making my own commit, I will add you as a contributor on my branch. I have modified it to compile inside a docker container or github actions, and made vulkan work from your version to the best of my knowledge properly.

I am also going to see if I can make the docker compilation not break the local comp of gamescope (due to path differences) (just make clean to fix) and the package cache for rust not being available currently as passed through, and thus downloading every time we compile with docker. I need to see if I can use overlay fs or something similar to avoid copying these large directories, because its not worth the extra disk space for a full clone of the project dirs into the docker image. If anyone has a great idea for this, let me know, but I will check later today to see what I can do.

I will be doing some small fixes in a bit, but I am open to comments!

Twig6943:
@davidawesome02-backup i'll give it a shot as soon as I can. So far everything seems good. Thanks for adding me as contributor :)

Twig6943:
@davidawesome02-backup seems good to me

Blahkaey:
Superseded by #172, Closing this one.

PR #130, opened 2026-01-11
Big picture mode UI

Blahkaey:
This feature adds a new ui that is designed to be used in steam big picture mode.

Launching partydeck with the arg --handler "handler name" will launch the app into this new ui. The handler is pre selected and the user just needs to add the players and devices before pressing start or enter to launch into the game. When running this ui through steams gamemode/bigpicture mdoe add in the --kwin arg

To add a player instance just press controller A button, right mouse, or Z. Once pressed a player  is added to the ui with the device used. The user can add additional devices to an player by clicking the plus button to add the next pressed device to that instance (just realized now that only a mouse can trigger adding additional devices to a instance).

A player can be removed by pressing controller B Button or backspace, a player instance can only be removed by the devices assigned to that instance.

The assigned profiles are sequentially picked from the list of profiles added to the app, if there is no profiles they are auto created. Player profiles can be changed by using the left and right arrows on the keyboard or left right dpad buttons on the controller, or by clicking the arrows buttons on a player with the mouse. Only the device assigned to a player can change a players profile with the mouse being the exception as it can click any players arrow buttons.

The ui is affected by desktop scaling if launched on the desktop, but if launched in gamemode with kwin desktop scaling is irrelevant.

I think that's all the features of this ui, im open to any suggestions if anyone has any.

https://github.com/user-attachments/assets/b102a96e-11f8-4099-84e1-a1a6b023f370

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -120,7 +120,7 @@ fn main() -> eframe::Result {
         None
     };
 
-    let fullscreen = std::env::args().any(|arg| arg == "--fullscreen");
+    let fullscreen = !auto_handler_id.is_empty() || std::env::args().any(|arg| arg == "--fullscreen");
 
     std::fs::create_dir_all(PATH_PARTY.join("handlers"))
         .expect("Failed to create handlers directory");


davidawesome02-backup:
Nice, I really like this idea, I think it could go great with my previous thoughts for a `--auto` mode to allow adding as a prefix to any steam game, but that is yet to be done. 

It looks like you will just keep the settings that were set in the main settings page (like to use kwin or gamescope, etc.) should we allow users to enter the main UI from here just, or vise-versa to do things like change those settings?

Also, reading through the code, I noticed a few things (untested yet) that I don't fully understand why they were done, mainly:
```rust
unsafe { std::env::set_var("XDG_CONFIG_HOME", PATH_PARTY.join("kwin")) };
```

Was this intended as a testing setting, or why is this here? I don't know if I am missing something, but if you are doing this for the kwin scripts, the PATH (wunnr just added to his version I think) should take care of that.

Lastly of the things I noticed, I don't think you currently have limits on the device name's length, and I didn't see any way to scroll / extend the view if you added too many boxes / are on a too small display size. These last two are nit-picky, and honestly don't matter much, but though you might want to know.

Blahkaey:
Hey thanks for the feedback. yeah i really enjoyed adding this feature as it makes the experience of playing split screen with someone else much more seamless when paired with steams big picture mode.

Yeah i did not mean to add the env XDG_CONFIG_HOME. I found it fixes an issue i was having related to kwin and desktop scaling. basically partydecks kwin inherits the desktop scaling (my display is 3840x2160 with a desktop scale of 200%) and causes ingame resolutions to be incorrect. By setting the XDG_CONFIG_HOME path partydecks kwin uses a different path for its configuration files preventing it from inheriting the desktop settings. i will add a pr for this later.

Regarding your notes, device names are shortened but just for mice and keyboards as i found them to have the longest names, i should probably just do it for all devices, somewhere in the render_device_row func im just using the first word of the device name. When you say add too many boxes do you mean player instances? if so there is a limit of 4 player instances at a time and yep definitely need to handle different screen sizes better.

I also like the idea of opening up the regular ui from this one and vice versa, i will look into that aswell

My current todo for this feature is:
- Change the arrow buttons to visually press when a controller is cycling the profiles
- Better handling of different screen sizes
- Instructions on how to use the ui - maybe add a toggle in the settings for users to turn it off for a cleaner display - or a hotkey to reveal input guide 
- Make device name shortening affect all device types
- Remove set XDG_CONFIG_HOME
- Allow rescanning of devices

wunnr:
Very cool! I do have some ideas that may improve the UX, will post more detailed feedback soon.

PR #134, opened 2026-01-18
Case unsensitive check for .exe and .bat (useful for old games like NFS Underground 2)

framilano:
I'm writing an handler for NFS Underground 2 and I had to rename the executable to speed2.exe instead of SPEED2.EXE to let it recognize as a Proton game.

### DIFF
--- a/src/handler.rs
+++ b/src/handler.rs
@@ -135,7 +135,10 @@ impl Handler {
     }
 
     pub fn win(&self) -> bool {
-        self.exec.ends_with(".exe") || self.exec.ends_with(".bat")
+        self.exec.ends_with(".exe") || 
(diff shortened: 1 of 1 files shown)


wunnr:
Good catch, thanks

davidawesome02-backup:
@wunnr while I think this change is good, I think this misses the `.cmd` file ext (same as `.bat` in windows) and all other cases other than just with or without full caps (like .Bat although most likely rare to see in a production game) ... I will see if I can make a commit in a few days (again more things to do lol) just so we have further compatibility, even if its not currently a problem :)

Thanks framilano!

framilano:
@davidawesome02-backup created a new PR (due to this one being closed)

https://github.com/wunnr/partydeck/pull/135

PR #96, opened 2025-11-16
Remove unneeded dependancies and implement nested compositor layouts in RIVER & KWIN

davidawesome02-backup:
NOTE: THIS IS FROM A SERIES OF FEATURES, this removes dependencies as well as implements river game layout, use just my `main` branch for the one that just removes unnecessary dependencies.

Implements river window layout management as well as nested kde layout after launch.

I copied / re-implemented my C version of the river game layout, accessible via a `--internal-layout` option that passes a FD to get nested virtual monitor information back at the host. When starting with either option (kde or river), we launch the compositor, then inside there, we launch a new version of river that sends back up its state. We then launch instances inside that `DISPLAY` / `WAYLAND_DISPLAY`. We watch all child processes, then when any are closed, we just check if it was the nested compositor and exit instantly or if its a game, count down until all games die.

New gamescope changes allow you to unbind the inputs via win+G like specified in the gamescope docs. Now gamescope builds support `--nested-follow-window-scale` allowing the games launched inside sessions to resize dynamically if one should die, or for better nested support.

Fixed a few layout bugs at small sizes with settings and the top bar being clipped on smaller screens (I use a 1600x900 screen, and it spawns small and is hard to use without adding this at least in settings).

Also, the big xml files are required for us to be able to talk to river (so we don't have to hardcode things) in case river ever changes. 

RIVER IS NOT REQUIRED TO BUILD OR RUN, IT IS FULLY OPTIONAL.

This is the main change I was interested in commiting as of lately; but I hope to over the next few days change a few other things mainly:
1. Launching should search PATH rather than using hard coded local paths, and should instead just offer a option to prepend our binary's path to the PATH variable.
2. The build script can be replaced with a more pure rust version via build.rs compiling the binaries, then copying them to `/src/bin` (excluded from git) where `cargo build -F build_all` can be made to allow us to build all the deps (and build_NAME) to build each of the different binaries.
3. Building off of 2, we can also use include_bytes (as well as a feature flag) to allow `-F embed_all` to allow directly embeding (and extraction at runtime) of gamescope, umu, gbe, and the kwin scripts to avoid the current .tar.gz export, and instead be able to supply a single binary.
4. Ability to directly add devices via a selection on the right hand side / using a path box where you can add it your self (my vm I use to test this splits my mouse into two mice where one clicks and one moves, making movement inaccessible to party deck by default)
5. Make players (via a setting disabled by default) able to be created without inputs bound to them, this is useful for testing (I have been doing this via directly disabling features in the code).
[...]

PR #141, opened 2026-02-13
Remove unnessisary dependencies - see #96

davidawesome02-backup:
Split from discussion on #96. Removes now unneeded dependencies we have added over time as well as small QOL changes.

Dependency changes
- Updated most dependencies to be `default-features=false` to remove all the extra added layers of deps.
- Removed `rand` package as it was redudant to have `rand` and `fastrand` installed
- Updated `image` and removed (unused unless someone added it into a handler) jpeg support.
- Explicit listing of egui backends we wish to use.
- Removed `semver` used for update checks - replaced with a simple string comparison as we expect any version change to come as a result of an update anyway.
- Removed SDL3 requirement replaced with x11 api directly (as used internally by sdl3)

QOL changes:
- Added option to not check for updates every time the program is launched.
- Fixed settings bug on small screens (would hide the lower bar) and didn't scroll if we had options hidden

Its pretty simple changes over all, but done to minimize the extra compile and download time.

This is the prerequisite to the next changes I have made, but if need be I can rebase those.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1858,6 +1858,8 @@ dependencies = [
  "num-traits",
  "png",
  "tiff",
+ "zune-core",
+ "zune-jpeg",
 ]
 
 [[package]]
--- a/src/monitor.rs
+++ b/src/monitor.rs
@@ -23,27 +23,52 @@ impl Monitor {
     }
 }
 
-// This should mimic the SDL monitor retrival used by gamescope, while avoiding all of SDL.
+// This should mimic the SDL monitor retrival used by gamescope, while avoiding all of SDL. (IGNORES SDL_HINT_VIDEO_DISPLAY_PRIORITY, and if display dosnt have "visual info" because all modern one will)
+// https://github.com/libsdl-org/SDL/blob/225fb12ae13b70689bcb8c0b42bf061120fefcc4/src/video/x11/SDL_x11modes.c#L868
 fn get_monitors_x11() -> Result<Vec<Monitor>, Box<dyn std::error::Error>> {
     let (con, screen_num) = x11rb::connect(None)?;
     let screen = &con.setup().roots[screen_num];
 
-    let res = con.randr_get_screen_resources(screen.root)?.reply()?;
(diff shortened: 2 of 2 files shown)


wunnr:
We may want to keep jpeg support, some of the handlers on the list I believe already use jpegs.

davidawesome02-backup:
@wunnr 
> We may want to keep jpeg support, some of the handlers on the list I believe already use jpegs.

Okay, added jpeg support back into it for now; let me know if you want any other changes for this commit.

wunnr:
Just tested multi-monitor; it appears that the SDL-less function for getting monitors does not return them in the same order that the current SDL-based function does, and therefore results in gamescope windows going to the wrong monitor. e.g. on current partydeck, running get_monitors_sdl() on my PC returns DP-1 (external monitor) as monitor 0 and eDP-1 (my laptop screen) as monitor 1, whereas running get_monitors_x11() on your build returns the opposite.

(The more I have to deal with multi-monitor headaches, the more I regret implementing this feature at all. But I digress...)

davidawesome02-backup:
> Just tested multi-monitor; it appears that the SDL-less function for getting monitors does not return them in the same order that the current SDL-based function does, and therefore results in gamescope windows going to the wrong monitor. e.g. on current partydeck, running get_monitors_sdl() on my PC returns DP-1 (external monitor) as monitor 0 and eDP-1 (my laptop screen) as monitor 1, whereas running get_monitors_x11() on your build returns the opposite.
> 
> (The more I have to deal with multi-monitor headaches, the more I regret implementing this feature at all. But I digress...)

@wunnr 


Okay, this SHOULD work now, I tried to copy the exact code (as commented in the commit) from SDL2 if this doesn't work on your system, ping me again, I tested as best I could on my system, and it matched with the sdl2 direct calls but at once point the gamescope param didn't work at all for me, just getting stuck on one monitor (would have affected both versions).

Sorry I didn't catch this earlier, I think it was just that the primary monitor is sorted to be first in the list, and you probably don't have your primary in the "correct" place in the normal list or ... something?

wunnr:
Seems like the change did indeed fix this, thanks!

PR #142, opened 2026-02-14
Add settings for auto selecting profiles when adding instances

zhliau:
I found that when playing local co-op, most often we'd want to continue from a saved game, so it got tiring having to always pick the profile to use. This PR adds the ability to allow auto-selecting profiles when adding controller instances before game start.

- If the priority list of profiles is exhausted, guest profiles will be used
- Config option added to enable auto selecting profiles by priority
- Priority order modifiable via UI buttons

image


Blahkaey:
Hey thanks for the contributions. sorry it took so long for anyone to get back to you. unfortunately im going to close this as we have two big ui updates incoming which will cover this feature

PR #153, opened 2026-03-08
build: added release workflow and moved build scripts to build.rs

Blahkaey:


davidawesome02-backup:
Hey @Blahkaey, reviewing this right now and I noticed a few things I think I would change before we merged this PR. One concern I had with this is the use of the https://crates.io/crates/sevenz-rust that when checking the [github repo](https://github.com/dyz1990/sevenz-rust), we find that the main developer has now entirely removed their account and thus all repos. I don't think that is a safe dependency to include in this project as such, even if just as a build time version. I found that https://github.com/hasenbanck/sevenz-rust2 still exists, but it appears to be a somewhat small fork off the original, and still has a bunch of dependencies that get added with how you have it here. I think we can disable most of the default features, as we don't need `default = ["aes256", "bzip2", "compress", "ppmd", "util"]` (we *may* need ppmd, but I don't even think that's required.)

Also, we may want to just make the build script actually compile from source the official repo instead of doing this downloading from releases. As a git submodule, we can add a rust subproject that has a build.rs to run the cmake crate in order to build this, or expose that as a build flag, to either build, download, or ignore. Currently you have this building gamescope as a separate phase in the github workflow, but we could just have it as a `cargo run -F build_gamescope`.

Not sure how you would want to implement this, but I personally prefer building the deps on the local machine, or offering a way to disable downloads should be added as a feature flag (can be defaulted to on) just to allow development with testing versions of those other repos.

Also because I am mentioning some dependency things, do you think we should keep `reqwest` using the `http2` feature? I don't see why we are using that currently and I think I forgot to remove before when removing the unused sub-projects.

`reqwest = { version = "0.12.15", features = ["blocking", "json"] }` -> `reqwest = { version = "0.12.15", features = ["blocking", "json", "charset", "default-tls", "system-proxy"], default-features = false }`

I can take care of that later in a separate commit, just thought I should point it out while I am checking all the dependencies.

Let me know how you feel about adding those build steps / options as features, because otherwise this generally looks fine. I will test the github workflow in my private repo in a few hours.

If anything here is unclear, let me know so I can better explain, but I generally think this is good, just not sure how you would want to deal with the deps fully.

Blahkaey:
Hey thanks for catching those issues, I've made the changes based on your feedback:
                                                                                                                                                                                                          
- Swapped out sevenz-rust for sevenz-rust2 with default-features = false and only the util feature enabled as its needed for the decompress func.
- Added download_deps default on and build_gamescope feature flags, so you can now do cargo run -F build_gamescope or build with --no-default-features to skip downloads entirely.
- Haven't touched the reqwest default-features yet since you mentioned doing that separately, but happy to include it in this PR if you'd prefer.
- For source building umu and gbe on linux would be doable, the problem is the windows gbe binaries. the windows dll need msvc on windows and the msys2/mingw path is marked as broken upstream due to ABI differences with the overlay library, so not really sure on what to do here.

Let me know what you think about the Windows gbe situation

Blahkaey:
The submodule glm in the gamescope repo seems to be broken to i changed the readme git clone command to avoid the command failing. Also moved downloaded dependencies to deps/releases/, which is already in .gitignore on older branches preventing untracked files from appearing when switching between branches.

davidawesome02-backup:
Hey @Blahkaey, did some updates for what I think we needed to do with this, I am not sure if you agree on how this should be done, but this adds a hash it checks to prevent dependency replacement later as an option, and adds some logging in here, fixing the accidental change you had that broke the compile.

See my branch: https://github.com/partydeck/partydeck/commit/eac36d9263478cff5f3cc74fa9ae7797e270a5e5

Blahkaey:
Thanks for the additions, they look good and will add it to the pr soon. If you want any other changes your welcome to push them to the pr

PR #160, opened 2026-03-29
Big mode UI

Blahkaey:
Will add notes later.. i need to goto sleep


Blahkaey:
closing this version of bigmode ui for now. will revisit the layout system idea at a later point

PR #161, opened 2026-04-01
Remove walkdir dependency, use Zip crate for zip_dir

Xyphyn:
This removes the walkdir dependency as it was only used in one place.

The zip_dir function now uses the already-installed zip crate instead of calling to shell commands.


Blahkaey:
looks good to me

PR #166, opened 2026-04-03
feat: add frame limiting and set gamescope refresh rate

Blahkaey:
Adds a configurable frame limiter. there is global options and override options per handler.

refresh rate can be set through the -r arg on gamescope, this fixes an issue where in some games because of the timing where game scope auto sets the refresh rate, the game does not pickup the correct number and defaults to 60fps

### DIFF
--- a/src/launch.rs
+++ b/src/launch.rs
@@ -184,32 +184,13 @@ pub fn launch_cmds(
             }
         }
 
-        let mut mangohud_config: Option<String> = None;
         let frame_limit = h.frame_limit_override.unwrap_or(cfg.frame_limit);
-
         let mangohud_framelimit = frame_limit > 0;
-        let use_mangohud = h.use_mangohud || mangohud_framelimit;
 
-        if use_mangohud {
-            let mut config_parts: Vec<String> = Vec::new();
-
-            if mangohud_framelimit {
-                let method = match cfg.mangohud_limit_mode {
-                    MangoHudLimitMode::Early => "early",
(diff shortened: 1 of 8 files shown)


src/launch.rs
```diff
@@ -233,6 +247,19 @@ pub fn launch_cmds(
...
+
+        if mangohud_framelimit {
+            println!("[partydeck] Frame limiting enabled");
+            let method = match cfg.mangohud_limit_mode {
+                MangoHudLimitMode::Early => "early",
+                MangoHudLimitMode::Late => "late",
+            };
+            let mangohud_config = format!("fps_limit={},fps_limit_method={},no_display", frame_limit, method);
+            cmd.args(["--setenv", "MANGOHUD", "1"]);
+            cmd.args(["--setenv", "LD_PRELOAD", "/usr/$LIB/mangohud/libMangoHud_shim.so"]);
```
davidawesome02-backup, line 259: This most likely should not overwrite LD_PRELOAD entirely, we want to save the environment's LD_PRELOAD, just append to it. Do you mind adding a getenv call and just prepending it with `:` so we don't cause problems?

This is the only real thing I see so far as a possible issue, I will try to compile and run some quick tests in a bit, but if this is fixed, I can merge it. Sorry I didnt review sooner, entirely forgot to come back after marking as read when I had limited time.

Blahkaey:
Closing this pr as parts of this work will be integrated into the compositor feature instead

Issue #163, opened 2026-04-02
Motivated and looking to build on COSMIC Support


specificocean23:
G'day guys, I'm working on adding COSMIC compositor support. From researching: the compositor exposes zcosmic_toplevel_manager_v1 v4 which should allow clean window management without KWin. I'm planning to add a CosmicBackend implementing the same interface as the KWin backend. Any guidance on the architecture or things to watch out for?

davidawesome02-backup:
@specificocean23 

Hey, nice to hear you are interested! We are currently in discussions actually about changing the way layouts are done, so you might want to wait a bit. My river commit adds a new system for launching nested sessions of both kwin and river, and sets up a system we will most likely be using into the future for nested situations if thats something you are targeting. I am just now looking into making a standardized way to interact with the compositors for window positioning in a less opinionated way (like having a base class that sends specific wayland api calls that follow a similar structure), but unfortunately looks like a landmine for most of.... well everything wayland related.

Were you thinking of the nested use case and / or the main compositor layout handling like we allow for KWin?
Do you have any thoughts on what this feature could add to the current backends if we are to support river wm as a nested mode or some other simple single executable nested mode compositor?

This isn't to say this wouldn't be a good feature, and I would say it should be fine to implement if we have a justification for it, I just personally want to avoid too much out of scope that may eventually get cut away due to restructuring or better ideas we are trying to take on right now.

I would love to get your input though, and any contributions would be helpful!

PR #149, opened 2026-02-21
Rework launching to support diffrent nested backends, specificly adding river support

davidawesome02-backup:
Implements wayland management for river support (and possible future similar nested compositors)
Adds my river-game-layout implementation for window layout inside of river.
Adds kwin nested durring launch option instead of the current kwin wrapper and relaunch (we should move to depreciate that)
Update my reference in the project to properly link to my account

[Already done, so being removed to resolve merge conflict] Adds new gamescope (with --grab) to allow users unbinding keyboard / mice inside of nested instances if necessary.
With new gamescope, we also enabled the `--nested-folow-window-scale` flag allowing games to die and gamescope to dynamically resize instead of staying at broken resolutions, and `--force-windows-fullscreen` to fix the games launching as floating windows (scale problems esp after resize event)

Explanation of changes specifically that I think need more spotlight (see #96):
Added `nix, wayland-client, wayland-scanner, wayland-backend` as listed dependencies (already were sub dependencies, so no real new packages are required) in order to allow us to talk to wayland API's for river setup, and use nix handle new game launch code.

To launch now, we first trigger a relaunch of our own executable with the flag `--internal-layout` which accepts a target file descriptor (for writing back display information used to launch games inside the nested display), and a target full width and height of the layout (added with the request to make river expand to fill full screen rather than default size); to test just this portion, run partydeck with `--internal-layout 0:1920:1080` inside of a river or nested compositor to see the output to the console, and trigger a resize / layout inside of river instances without using partydeck fully.

Launch collection code moved away from just waiting for all programs to exit, now exiting early if all games die, or all compositor die (just in case games or river die without closing each other, as river dosn't exit when games die by default, all are killed fully if still alive on exit condition)

ALL XML files added are just upstream wayland protocols as far as I could find the latest versions.

Please let me know if you have any problems with this, I know its pretty big, but I hope the addition of nested compositors will allow us to have more flexibility for users without kwin (or just not wanting to affect the full desktop)

PR #170, opened 2026-04-12
hyprland automatic layout and window sizing

jonboh:
Hi! I'm not sure if you are willing to accept changes to support other desktop environments apart from KDE out of the box, if you are not interested it is completly ok, feel free to close this PR in that case.
Otherwise, I've implemented the logic from kwin into hyprland using hyperctl.
I discovered `partydeck` last week and I haven't had the time to test it a lot, but as part of configuring my rig to run it I ended up building this functionality.
As far as I've test it it works well, there are only two caveats, I've only tested it running Broforce (which starts up quickly), I have still to try it out with a game that takes longer to load and start up. And the other thing is that I seem to have something misconfigured in my system that results in gamescope ignoring `--display-index` and all game instances are launched on the display in which I have the focus. From the logic implemented I think the layout should be correctly handled when the games are spawned in the correct window, but I've been unable to test this out. This is not critical for my use-case as I only have one monitor in which I want to run all game instances.

Let me know if this is a feature you are interested in, and if there is something in the implementation you'd like to see changed.

Apart from that, I've noticed that there are a number of lints that are flagged by `clippy`, have you considered adding it to the development? I could make another PR with those fixed.

### DIFF
--- a/src/app/app_pages.rs
+++ b/src/app/app_pages.rs
@@ -499,6 +499,14 @@ impl PartyApp {
             self.infotext = "DEFAULT: Enabled\n\n Resizes/repositions instances to fit the screen using a KWin script. If using a desktop environment or window manager other than KDE Plasma, uncheck this; note that you will need to manually resize and reposition the windows.".to_string();
         }
 
+        let enable_hyprland_check = ui.checkbox(
+            &mut self.options.enable_hyprland_windows,
+            "(Hyprland) Automatically resize/reposition instances",
+        );
+        if enable_hyprland_check.hovered() {
+            self.infotext = "DEFAULT: Disabled\n\nResizes/repositions instances to fit the screen using Hyprland IPC. Enable this when running on Hyprland. Requires hyprctl to be in PATH.".to_string();
+        }
+
         ui.horizontal(|ui| {
             let split_style_label = ui.label("Split style");
             let r1 = ui.radio_value(
--- a/src/app/config.rs
+++ b/src/app/config.rs
@@ -22,6 +22,8 @@ fn default_true() -> bool {
 pub struct PartyConfig {
     #[serde(default = "default_true")]
     pub enable_kwin_script: bool,
+    #[serde(default)]
+    pub enable_hyprland_windows: bool,
     #[serde(default = "default_true")]
     pub gamescope_fix_lowres: bool,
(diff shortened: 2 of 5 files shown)


davidawesome02-backup:
@jonboh 
Hey, great to see new contributions! We are currently actually about to rewrite some of the launching logic to allow for more flexible window managers to be added! See #96 or #149 for my changes adding RiverWM support (and a reworked launch code) as a backend. Also we have been asked to add Cosmic support in #163. We expect to merge river in the next few days ( Unless we change our minds again :P ) and we are discussing how we want to do the many window managers into the future.

I see you are using hyprlandctl here to control the window placement / do layout for the added windows - Do you know if this is something that can be done over a wayland api (I think this is what hyprlandctl is doing, as most are just wayland API's)? If it is, we can most likely add that into the same river code, so we have less sprawl for the nested managers. (we also intend to soon figure out how kwin can be controlled by the rust program rather than separately as we currently do.)

My River support also adds a new layout option that I would hope to support soon in kwin or any other added window managers, while trying to keep more of the layout engines inside the base project. 

Do you know the status of hyprland with HDR or high refresh rates. River is what I chose to support primarily due to the ability to easily do custom layouts and the fact it can be distributed as a single executable small compositor, but does not support either of these two nonstandard options.

Of note, currently the river code is only for nested mode, but I intend to add more general non-nested operation and implement kwin more directly inside that same code to simplify the codebase.

Sorry you caught us at a time where we are going through a partial rewrite with the launch code. As long as hyprland does not get too complex, I don't mind merging as a nested or normal mode manager in the coming updates, but I do think we should hold off on this specific implementation so we can have a more normal approach.

If you want to get more updates on the current project plans, feel free to join us on the new discussion chat we just started on matrix.

Thanks again!

jonboh:
makes sense, I wouldn't want to have to support each desktop environment independently if I maintained the project. I would love to give you more insight into hyprland/wayland but unfortunately I'm not that familiar with this environment. I only use it on my game streaming setup, my main machine still runs x11/i3. I used the opportunity to check out hyprland, as I've heard good things about it, but I'm not familiar with the internals needed to interface with wayland through an API or its status regarding HDR.

I'll wait for your merge and if I get a chance and some extra time I'll revisit this :)
Thanks for your work on this project!!

PR #176, opened 2026-05-05
Controller connection management & graceful shutdown

OJUrbaniak:
Better Controller Support
- We manage inputs by creating virtual gamepads. These now copy your real controller's name and ID exactly. This fixes issues where buttons or triggers didn't work in-game.
- Hotplugging: You can now unplug and plug your controller back in while playing. 
- The virtual gamepad will automatically re-grab the physical device.
- We use background threads that wait for the exit signal. To my understanding this is probably the most performant way of going about this.

How to Test
   1. Launch any game in split-screen.
   2. Verify that all controllers move the correct characters.
   3. Unplug a controller and plug it back in; verify it still works.
   4. Close the game and verify the UI immediately returns to the main menu and cleans up all temporary folders.

In the interest of transparency - I am not a Rust dev (I mainly use java/php/ts) so most of the code is generated. It works perfectly for me but I would appreciate some feedback. I've attached a video and image of it working.

Thank you for all the hard work that you put into maintaining the repo!

https://github.com/user-attachments/assets/a46828d7-21f5-448c-b0ca-6c6549c24608

Screenshot_20260505_215956

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,5 +1,6 @@
-use std::{fs, path::PathBuf};
-use std::path::Path;
+use std::fs;
+#[allow(unused_imports)]
+use std::path::{Path, PathBuf};
 
 #[cfg(all(not(feature = "download_deps_latest"), feature = "download_deps"))]
 use sha2::{Sha256, Digest};
--- a/src/app/app.rs
+++ b/src/app/app.rs
@@ -1,4 +1,5 @@
 use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::mpsc::{channel, Receiver, Sender};
 use std::sync::Arc;
 use std::thread::sleep;
 
@@ -13,7 +14,7 @@ use crate::util::*;
 
 use eframe::egui::{self, Key};
 
-#[derive(Eq, PartialEq)]
+#[derive(Eq, PartialEq, Clone, Copy)]
 pub enum MenuPage {
     Home,
     Settings,
@@ -21,6 +22,7 @@ pub enum MenuPage {
     EditHandler,
     Game,
     Instances,
+    Active,
 }
 
 #[derive(Eq, PartialEq)]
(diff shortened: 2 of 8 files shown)


davidawesome02-backup:
@OJUrbaniak Good to hear you are interested in the project!

Unfortunately I cannot merge this PR due to a few issues:
1. Blahkaey, Wunnr, and I have not decided on the use of fully AI generated code, and honestly I don't know if that will be legal with our current license.

2. You removed multiple existing features here that we actually do use currently, and we cant just break things lol

3. The code attempts to use uinput to create a virtual controller input device, and while this works, any controllers with special drivers or options not expressible via uinput (like rumble) are NOT supported with this path easily, and not implemented here.

4. uinput requires that you have a extra kernel module loaded, and have it setup to allow non-root users to access (not the default, but many applications including steam set that up by adding you to the `input` group), still requires extra permissions that are not really needed.

A few things of note as well:

We right now use the least CPU intensive version of waiting for processes to exit, just has a small bug with the waiting screen. We use the `wait` syscall that just legitimately waits for the process to exit before the kernel will let us continue. This code has been refactored in my riverwm program to be also minimally impactful, but allow us to kill the games. From what I see, you tried to move this to a input handler subsystem that you made to handle forwarding of inputs until all games die, and this works for this type, but is somewhat out of place for where game death is handled.

I like the thought of hot plugging, but the way its done here unfortunately is just not going to be possible to use currently, just due to the problems this approach has.

I really need to make a commit to format the code soon, because a lot of the changes are just from you running the formater lol

And if you do have problems, like you note of some controllers not working in game, please let us know, we are open to different problems, and while this fix for it isn't something we can accept, we can attempt to add some of the small patches you appear to have added if those fix the bug directly:
```
cmd.env("PROTON_DISABLE_HIDRAW", "1");
cmd.env("WINEDLLOVERRIDES", "xinput1_3=n,b");
```

Thanks so much for being interested in the project, I am sorry that I have to close this. I don't intend this as a disapproval of any code being merged, or of your ability to add some of these changes, just this PR is not something that meets the current project direction. Please feel free to open changes if you want to learn rust on this project, (like I am trying to do still), I hope this can be something you can be able to have fun with into the future, using it for games or to practice programing \:P

Feel free to respond here or on matrix if you want!

Issue #179, opened 2026-05-17
[Game issue] (Necesse): Not Launching


FelipeFTN:
- [x] I have already tried adding my game through the wizard or downloading an existing handler package and my game does not work or is problematic

**PCGamingWiki Article**
https://www.pcgamingwiki.com/wiki/Necesse

**Additional info**
Just did the default setup:
- Build **partydeck** locally with no issues
- Run with `cargo run` 
- Added the Game from the Wizard (with defaullt configuration)
- Game doesn't launches.

I have tried the pre-compiled binaries from Releases tab. Didn't work too.
I have tried different configurations for SDL2 Override and Linux Runtime; nothing seems to work.
Openning _Necesse.jar_ just by double clicking it works nicely. The error seems to be only when trying to open it with partydeck.

**Logs**
```
[partydeck] Instances:
[partydeck] Creating profile .Gort
[partydeck] Profile created successfully
[partydeck] Creating game save Necesse for .Gort
[partydeck] Profile save data created successfully
[partydeck] - Profile: .Gort, Monitor: 0, Resolution: 1920x1080
[partydeck] INSTANCE 1:
[partydeck] CWD=/mnt/SSD/SteamLibrary/steamapps/common/Necesse
[partydeck] ENABLE_GAMESCOPE_WSI=0
[partydeck] SDL_GAMECONTROLLER_ALLOW_STEAM_VIRTUAL_GAMEPAD=1
[partydeck] SDL_JOYSTICK_HIDAPI=0
[partydeck] "gamescope"
[partydeck]  "-W" "1920" "-H" "1080" "--"
[partydeck] "bwrap" "--die-with-parent" "--dev-bind" "/" "/" "--tmpfs"
[partydeck] "/tmp"
[partydeck] "/mnt/SSD/SteamLibrary/steamapps/common/Necesse/Necesse.jar"
[partydeck] ---------------------
[partydeck] util::kwin_dbus_start_script - Loading script /home/felipeftn/Desktop/partydeck/target/debug/res/splitscreen_kwin.js...
[partydeck] util::kwin_dbus_start_script - Script loaded. Starting...
[partydeck] util::kwin_dbus_start_script - KWin script started.
[gamescope] [Info]  console: gamescope version 3.16.23+ (gcc 15.2.1)
No CAP_SYS_NICE, falling back to regular-priority compute and threads.
Performance will be affected.
[...]

PR #180, opened 2026-05-17
Correctly detecting compiled java files

FelipeFTN:
The `.jar` file is being executed directly as a shell script instead of being run with `java -jar`. The system is trying to execute the JAR file as if it were a shell script.

When launch.rs detects a `.jar` file as the executable, it should prepend `java -jar` to the command.

Linked to #179

This PR successfully solves two issues:
- Java games like Necesse not even opening
- Very specific build error with `git apply` that I was facing

Now partydeck can open the Necesse game, but setting up windows for two controllers/players is not working properly; and I don't have any idea why and how to keep going from here.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,6 @@
 /deps/releases
 /deps/gbe-linux
 /deps/gbe-win
-/deps/gamescope
 /deps/umu
 res/gamescope
 gen_release.sh
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 /deps/releases
 /deps/gbe-linux
 /deps/gbe-win
+/deps/gamescope
 /deps/umu
 res/gamescope
 gen_release.sh
--- a/build.rs
+++ b/build.rs
@@ -81,11 +81,29 @@ macro_rules! build_println {
 
 #[allow(dead_code)]
 fn apply_patches(deps_dir: &std::path::Path) {
-    let mut git_apply = std::process::Command::new("git");
-    git_apply.args(["apply", &deps_dir.join("deps.patch").to_string_lossy()]);
-    let _ = git_apply.spawn().map_err(|e| {
-        build_println!("Failed to git apply the patches we have for our deps, this is most likely not a real error: {:?} - {:?}", git_apply.get_program().to_string_lossy(), e);
-    });
+    let patch_path = deps_dir.join("deps.patch");
+
+    // Check if the patch can be applied before attempting it.
+    // If the check fails the patch is most likely already applied, so skip silently.
+    let check = std::process::Command::new("git")
+        .args(["apply", "--check", &patch_path.to_string_lossy()])
+        .output();
+
+    match check {
+        Ok(output) if output.status.success() => {
+            let status = std::process::Command::new("git")
+                .args(["apply", &patch_path.to_string_lossy()])
+                .status();
+            match status {
+                Ok(s) if s.success() => { build_println!("Applied patches from deps.patch"); }
+                Ok(s) => { build_println!("git apply exited with {s} for deps.patch"); }
+                Err(e) => { build_println!("Failed to run git apply: {e}"); }
+            }
+        }
+        _ => {
+            build_println!("Skipping deps.patch (already applied or does not apply cleanly)");
+        }
+    }
 }
 
 fn main() {
@@ -135,7 +153,7 @@ fn build_gamescope(deps_dir: &Path, target_dir: &PathBuf) {
     let gamescope_dir = deps_dir.join("gamescope");
     let build_dir = PathBuf::from(std::env::var("OUT_DIR").unwrap()).join("gamescope-build-gcc");
 
-    if !build_dir.exists() && gamescope_dir.exists() {
+    if !build_dir.join("build.ninja").exists() && gamescope_dir.exists() {
         build_println!("Running meson setup command for gamescope");
         let status = Command::new("meson")
             .arg("setup")
--- a/src/app/config.rs
+++ b/src/app/config.rs
@@ -18,6 +18,10 @@ fn default_true() -> bool {
     true
 }
 
+fn default_kbm_support() -> bool {
+    pathsearch::find_executable_in_path("gamescope-kbm").is_some()
+}
+
 #[derive(Serialize, Deserialize, Clone)]
 pub struct PartyConfig {
     #[serde(default = "default_true")]
@@ -28,7 +32,7 @@ pub struct PartyConfig {
     pub gamescope_sdl_backend: bool,
     #[serde(default)]
     pub gamescope_force_grab_cursor: bool,
-    #[serde(default = "default_true")]
+    #[serde(default = "default_kbm_support")]
     pub kbm_support: bool,
     #[serde(default)]
     pub proton_version: String,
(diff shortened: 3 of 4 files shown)


build.rs
```diff
@@ -81,11 +81,29 @@ macro_rules! build_println {
 
 #[allow(dead_code)]
 fn apply_patches(deps_dir: &std::path::Path) {
-    let mut git_apply = std::process::Command::new("git");
-    git_apply.args(["apply", &deps_dir.join("deps.patch").to_string_lossy()]);
-    let _ = git_apply.spawn().map_err(|e| {
-        build_println!("Failed to git apply the patches we have for our deps, this is most likely not a real error: {:?} - {:?}", git_apply.get_program().to_string_lossy(), e);
-    });
+    let patch_path = deps_dir.join("deps.patch");
```
davidawesome02-backup, line 84: Yep, I see how adding the check before can be good. I added this, and choose to just ignore the error because its only something that runs when compiling gamescope, and it not applying, or failing, basically failure mode is not really a problem for this function. That is why I choose to just have it display "most likely not a real error". Because upstream gamescope got fixed recently, I may fix this soon, or just remove it entirely because the patch will be empty.

build.rs
```diff
@@ -135,7 +153,7 @@ fn build_gamescope(deps_dir: &Path, target_dir: &PathBuf) {
     let gamescope_dir = deps_dir.join("gamescope");
     let build_dir = PathBuf::from(std::env::var("OUT_DIR").unwrap()).join("gamescope-build-gcc");
 
-    if !build_dir.exists() && gamescope_dir.exists() {
+    if !build_dir.join("build.ninja").exists() && gamescope_dir.exists() {
```
davidawesome02-backup, line 156: This might be good to add in case you `rm -rf *` inside the directory, but honestly I just never did this \:P Is that what happened for you, or is there any other reason this caused problems?

I think we can add this if we want. Have you tested if this .exists works the same if the top parent directory is removed too? (It should, but I will need to test)

.gitignore
```diff
@@ -4,8 +4,9 @@
 /deps/releases
 /deps/gbe-linux
 /deps/gbe-win
+/deps/gamescope
```
davidawesome02-backup: This should not be done upstream as gamescope should be kept in the worktree as a full downloaded dep, just to make it stay as a subproject always. The reason I assume you added this was because of the patch dirtying the gamescope tree, but as long as you dont commit to gamescope first, `git add` shouldn't add the gamescope dirty changes.

src/app/config.rs
```diff
@@ -18,6 +18,10 @@ fn default_true() -> bool {
     true
 }
 
+fn default_kbm_support() -> bool {
```
davidawesome02-backup, line 21: Sure, we can add this if it makes it better durring build times. For me I always have this in my path because I build always with `-F build_gamescope` first, but this is good for new people building.

src/launch.rs
```diff
@@ -358,6 +358,9 @@ pub fn launch_cmds(
         }
 
 
+        if path_exec.extension().map_or(false, |ext| ext == "jar") {
```
davidawesome02-backup: I don't suggest we change this here, instead I would say to use a .sh file or similar. Adding handlers for the many scripts and different environments you may have I think gets a bit out of scope, and should be handled in the next version of the handler system Blahkaey is making.

davidawesome02-backup (commented):
If you want, I think this would be good to merge the small fixes with the removal of .jar section (see comment) and removal of the .gitignore changes.

I only don't want to have the .jar handling because I feel its not quite what we want, and it definitely should be something done in the new handler system (so properly configurable for each handler), but with the current system I think this is more of a hack unfortunately. Also, .jar can be ran by many diffrent java installs on the system, and most games ship the version of java you have to run the game with. Using the system installed one is not ideal for these, and why I think a .sh is more "correct".

If we are making this small fixes PR, I think I am fine with checking off on a merge, and for that you might as well throw in removing the accesskit setting on egui (in the Cargo.toml) that I meant to before lol, its a dependency that only actually works on windows.

Sorry if I come off as harsh, I don't intend to. Thanks for pointing out the problems here, and please feel free to suggest any changes you think are good. We really do want new ideas for this, I only think this can be accomplished better with the new system. If this is a issue you really think is important we can look into it more.

build.rs
FelipeFTN: A previous build I made just failed and left the gamescope-build-gcc directory without a build.ninja (since meson setup was interrupted and didn't generate the build.ninja). So I was getting some errors when trying to run it. 
So the code now checks `!build_dir.exists()` on **build.ninja** to decide whether to run meson setup — since the dir exists, meson is skipped, so can't find build.ninja.

FelipeFTN: Btw the build failed because of a lack of some dependencies like `wayland-protocols` or `umu-launcher` — I can't remember quite well now XD

FelipeFTN:
The Jar file change are not necessary if I download a version of Necesse with .exe file — it also skips the need of a .sh file in order to run some jar file too.

This PR seems a bit useless to me now, since the main idea (to run Jar files correctly is dead).
I will keep testing stuff in order to make Necesse run properly and I will be adding the commits here, so I will keep the Pull Request Open and I will ask your review again as soon as I get some good results with Necesse :)

src/launch.rs
```diff
@@ -468,4 +464,4 @@ pub fn fuse_overlayfs_mount_gamedirs(
...
.../0x0110,0x22ba/0x1020,0x046d/0xc219,0x046d/0xc216,0x046d/0xc216,0x046d/0xc219,0x046d/0xc218,0x046d/0xc211,0x24c6/0x892b,0x24c6/0x892a,0x24c6/0x891a,0x0738/0x5266,0x0738/0x3384,0x0738/0x3480,0x0738/0x8818,0x0078/0x0006,0x045e/0x000e,0x045e/0x0285,0x045e/0x0289,0x045e/0x0289,0x20d6/0x0dad,0x146b/0x0c01,0x0810/0xe501,0x0955/0x7214,0x0955/0x7214,0x124b/0x4d01,0x1345/0x3008,0x0079/0x1843,0x0079/0x1844,0x057e/0x2019,0x057e/0x2019,0x057e/0x201e,0x057e/0x2017,0x057e/0x2017,0x057e/0x2017,0x057e/0x0306,0x057e/0x0330,0x057e/0x0306,0x050d/0x0803,0x2836/0x0001,0x2836/0x0001,0x045e/0x0202,0x11ff/0x3341,0x0e8f/0x0003,0x054c/0x0cda,0x0f30/0x1112,0x2c22/0x2012,0x2c22/0x2010,0x1532/0x0402,0x1532/0x0705,0x1532/0x0900,0x1532/0x0900,0xf000/0x0003,0x0079/0x0011,0x1a34/0x0809,0x7545/0x1122,0x06a3/0xf623,0x06a3/0xff0c,0x06a3/0x040c,0x06a3/0x0109,0x06a3/0x040b,0x06a3/0xf518,0x16c0/0x0487,0x28de/0x11fc,0x0111/0x1431,0x0111/0x1419,0x6666/0x8804,0xf000/0x00f1,0x044f/0xb320,0x044f/0xb323,0x044f/0xb300,0x044f/0xd009,0x044f/0xd008,0x12bd/0xd015,0x14d8/0xcd07,0x0079/0x0011,0x05ac/0x3232,0x0c45/0x4320,0x2717/0x3144,0x16c0/0x05e1,0x6666/0x0667,0x0583/0x2060,0x07b5/0x0315,0x289b/0x0080,0x289b/0x0003,0x289b/0x0060,";
```
FelipeFTN, line 471: Just new_line removal at the end of the file made by my autoformatter.

PR #183, opened 2026-05-23
Add per-handler toggle to keep hidraw enabled for PS controllers (fixes #87)

### LINKED ISSUE #87: [Bug] PlayStation controllers don't work on Windows games with Unity's new Input System
Not really a partydeck bug but I'll put it here to track if I make any progress on a real solution for this.

Unity games using the [new Input System](https://docs.unity3d.com/Packages/com.unity.inputsystem@1.14/manual/index.html) don't detect PlayStation controllers when run through Proton; this is likely because, according to my understanding, the input system accesses controllers directly through HID instead of using an API like SDL or Xinput. Partydeck uses an environment variable `PROTON_DISABLE_HIDRAW=1` so that PS controllers get translated to Xinput through SDL, but this might be the cause of the issue.

The current workaround is to use Steam Input for your PS controllers and set Partydeck's controller filter to `Only Steam Input`.

cseelhoff:
Fixes #87.

## Problem

Partydeck unconditionally sets `PROTON_DISABLE_HIDRAW=1` and `SDL_JOYSTICK_HIDAPI=0` for every game launch (`src/launch.rs`). This was originally added to prevent PlayStation controllers from appearing twice (once as XInput via hidraw, once as DInput via evdev). The tradeoff is that wine's `winebus` cannot translate the controller to XInput, so games that only enumerate XInput — notably Unity titles using the new Input System, such as V Rising — see no controllers at all.

Reproduced with a DualShock 4 + V Rising under GE-Proton: pad works in the partydeck UI, gets no input in-game.

## Fix

Add an opt-in per-handler `enable_hidraw_ps: bool` (defaults `false`, so existing handlers are unchanged). When set on a Windows handler, the two env vars are skipped, letting wine expose the PS controller as XInput via hidraw and letting SDL's HIDAPI backend map it.

Exposed as a checkbox on the handler edit page, shown only for Windows handlers (next to SDL2 override / Linux runtime — mirrors the existing per-handler pattern from 8402c59).

## Verified

- Builds cleanly (`cargo check` against upstream main, `nix build` of full package).
- With toggle **off** (default): launch env shows `PROTON_DISABLE_HIDRAW=1` and `SDL_JOYSTICK_HIDAPI=0`, matching previous behavior.
- With toggle **on**: env vars are absent and DualShock 4 input works in V Rising.
- Existing handler JSON files load without modification thanks to `#[serde(default)]`.

## Why per-handler instead of per-controller

The env vars are process-scoped, not device-scoped — wine can't selectively enable hidraw for one pad. The right tradeoff (DInput-only vs XInput-via-hidraw) is determined by the game's input API, which is a handler property.

## Diff size

3 files, +13 / -2.

### DIFF
--- a/src/app/app_pages.rs
+++ b/src/app/app_pages.rs
@@ -258,7 +258,7 @@ impl PartyApp {
         }
 
         if h.win() {
-            ui.checkbox(&mut h.enable_hidraw_ps, "Enable hidraw for PS controllers (needed for Unity new Input System games; may cause double input in some games)");
+            ui.checkbox(&mut h.enable_hidraw, "Enable HIDraw for non-Xbox controllers (fixes Unity Input System games; may cause double input in non-Unity games!)");
         }
 
         if !h.win() {
--- a/src/app/app_pages.rs
+++ b/src/app/app_pages.rs
@@ -258,7 +258,7 @@ impl PartyApp {
         }
 
         if h.win() {
-            ui.checkbox(&mut h.enable_hidraw_ps, "Enable hidraw for PS controllers (needed for Unity new Input System games e.g. V Rising; may cause double input in some games)");
+            ui.checkbox(&mut h.enable_hidraw_ps, "Enable hidraw for PS controllers (needed for Unity new Input System games; may cause double input in some games)");
         }
 
         if !h.win() {
--- a/src/app/app_pages.rs
+++ b/src/app/app_pages.rs
@@ -257,6 +257,10 @@ impl PartyApp {
             });
         }
 
+        if h.win() {
+            ui.checkbox(&mut h.enable_hidraw_ps, "Enable hidraw for PS controllers (needed for Unity new Input System games e.g. V Rising; may cause double input in some games)");
+        }
+
         if !h.win() {
             ui.horizontal(|ui| {
                 ui.label("Linux Runtime:");
--- a/src/handler.rs
+++ b/src/handler.rs
@@ -47,6 +47,8 @@ pub struct Handler {
     #[serde(default)]
     pub use_mangohud: bool,
     pub use_goldberg: bool,
+    #[serde(default)]
+    pub enable_hidraw_ps: bool,
     pub steam_appid: Option<u32>,
 
     pub game_null_paths: Vec<String>,
@@ -75,6 +77,7 @@ impl Default for Handler {
 
             use_mangohud: false,
             use_goldberg: false,
+            enable_hidraw_ps: false,
             steam_appid: None,
 
             game_null_paths: Vec::new(),
--- a/src/handler.rs
+++ b/src/handler.rs
@@ -48,7 +48,7 @@ pub struct Handler {
     pub use_mangohud: bool,
     pub use_goldberg: bool,
     #[serde(default)]
-    pub enable_hidraw_ps: bool,
+    pub enable_hidraw: bool,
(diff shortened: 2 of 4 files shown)


wunnr:
Your description of the problem isn't entirely correct: winebus through evdev _does_ translate PlayStation controllers to Xinput, while hidraw exposes them natively as Dinput controllers. This can be verified by running `wine control joy.cpl` with hidraw disabled: PS controllers will show up as XInput devices. You can also verify this with partydeck, any game that only uses XInput (no Dinput, not Unity) will work just fine with PS controllers. Rather than using XInput, Unity's GameInput API enumerates devices its own way, and however it's doing that specifically does not play nice with PlayStation-as-XInput fake devices. 

All that being said, if this PR does fix Unity games and doesn't otherwise interfere with the controller filtering done by bwrap we can look into testing it

davidawesome02-backup:
Thanks for the PR @cseelhoff! Good to see new faces work on the project. 

I unfortunately don't have any controllers (other than a Nintendo switch joycon set) to test with, so I cant really test this at the moment.

The code looks generally fine to me, I will have to check [if this second](https://github.com/partydeck/partydeck/pull/183/changes#diff-b28dbadb7b0fb54a6a72860d79ba4fa2c1228788e81a14dbd99f8536f5e317a9R206) `.flatten` is really required, but otherwise I don't see anything that is wrong here. 


@Blahkaey do you possibly have any playstation controllers to test with? 

@wunnr
> Rather than using XInput, Unity's GameInput API enumerates devices its own way, and however it's doing that specifically does not play nice with PlayStation-as-XInput fake devices.

Do you know if this affects any controllers other than playstation's controller set on unity games or the like? 

If that is correct I would not object to adding an extra option for disabling or enabling the launch argument configurations we allow for.


While I don't have any problems with the code, I would encourage the use of your own code and PR messages should you have the ability rather than using AI, but this will not cause me to object to the inclusion of this code currently.

Blahkaey:
Unfortunately I do not have a playstation controller to test with, I only have 360 and xbone controllers. 

The pr looks fine and touches only what is required for the feature. since its a toggle that only a specific group of users will use, this likely will be fine to add if it does reliably fix the issue with unity games. 

Regarding ai use, im fine with it as long pr authors who use ai actually review the code and make sure the changes are limited to what's actually needed for the feature

@davidawesome02-backup what do you think about a gh workflow that we can manually run on a pr so we can get a built binary that users in the discussion forum can test?

davidawesome02-backup:
Hey @Blahkaey, yes I think the code is acceptable, and I don't have too much of an issue with it, I just like to encourage people to write their own code when possible, not that tested and confirmed code written by AI is entirely bad.

I can work on a PR based build process in a little bit. The main thing is just going to be making sure we don't have an open attack vector like some projects have had with the PR's having too large a scope lol.

I am fine with the changes here right now, so I would be good with you merging assuming you are okay with it as well.

wunnr:
@davidawesome02-backup We should probably write up a contributor.md file explaining what sort of AI use is acceptable as well as general contributing guidelines.

I have several PS controllers so I'll get to testing this today to make sure it works and doesn't affect any of the regular controller filtering before we merge. Looking at the changes I'm already assuming it won't break anything but you can never be too sure. 

I believe the unity gameinput issue also affects Nintendo controllers but I'll test to make sure.

davidawesome02-backup (approved):
LGTM

PR #185, opened 2026-06-05
Per-instance render resolution, KWin multi-monitor placement, and audio sinks

anishgoyal1108:
## What

Three related per-instance "placement" features for splitscreen sessions:

### Per-instance render resolution + arg-token substitution
- Each instance gets an optional render-resolution override: gamescope renders the game at `-w/-h` and upscales it to the monitor output (`-W/-H`). Useful for weak-GPU downscaling and odd panels (e.g. 16:10). Exposed as a monitor-icon toggle on the Instances page.
- Argument tokens (`$WIDTH`, `$HEIGHT`, `$RESOLUTION`, `$PROFILE`, `$GAMEDIR`, …) are now substituted as **substrings**, so a token can be embedded inside an argument, e.g. `-ResX=$WIDTH` → `-ResX=1600` (needed by some Unreal Engine games). Path tokens (`$GAMEDIR`/`$HANDLERDIR`) pass through verbatim; every other arg is still run through `sanitize_path()`.

### KWin multi-monitor placement (+ classic fallback)
- The KWin splitscreen scripts now assign each instance to its chosen monitor by baking a per-instance assignment array into the script at launch, so multi-monitor placement works on the **Wayland** backend (where `--display-index` doesn't apply), not only the SDL backend.
- A new `kwin_multimonitor` setting toggles between this and **classic single-screen splitscreen** (ignore assignments, split whatever screen KWin opens windows on — the original behavior).
- The Wayland backend is only forced when the KWin script is enabled (it's what does the placement); otherwise gamescope picks its own backend.

### Per-instance audio sinks
- Each instance can route its audio to a chosen PipeWire/PulseAudio sink (`PULSE_SINK`), with buttons to create/remove/refresh `PartyDeck-N` virtual sinks (for routing players to different outputs via a patchbay). `pactl` errors are surfaced; sink naming uses the lowest free index.

Also: new instances auto-assign across available monitors, and `tmp/null` is recreated each launch so `game_null_paths` directory masking works.

## Testing
`cargo check` + `cargo test` clean.

### DIFF
--- a/res/splitscreen_kwin.js
+++ b/res/splitscreen_kwin.js
@@ -1,106 +1,77 @@
-const x = [
-  [],
-  [0],
-  [0, 0],
-  [0, 0, 0.5],
-  [0, 0.5, 0, 0.5]
-]
+// Loaded by PartyDeck. Two tokens are replaced at launch:
+//   PARTYDECK_ASSIGNMENTS      -> array mapping instance index -> screen index,
+//                                 e.g. [0, 1].
+//   PARTYDECK_USE_ASSIGNMENTS  -> true  : place each instance on its assigned
+//                                         screen (multi-monitor);
+//                                 false : classic splitscreen — ignore
+//                                         assignments and split whatever screen
+//                                         KWin opened each window on.
+// Instances sharing a screen are split (horizontal: stacked top/bottom).
+const assignments = PARTYDECK_ASSIGNMENTS;
+const useAssignments = PARTYDECK_USE_ASSIGNMENTS;
 
-const y = [
-  [],
-  [0],
-  [0, 0.5],
-  [0, 0.5, 0.5],
-  [0, 0, 0.5, 0.5]
-]
+const x = [[], [0], [0, 0], [0, 0, 0.5], [0, 0.5, 0, 0.5]];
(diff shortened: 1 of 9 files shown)


davidawesome02-backup:
Thanks so much for your contribution! 

All of these ideas are actually planned currently! That being said, the main reason they have not yet been implemented by us is due to the new rewrites currently going on. Unfortunately all of the changes listed here heavily conflict with the new changes. The main issues being in our new UI rewrite and moving to a new layout manager system entirely (currently considering dropping kwin and switching to a custom patched cosmic-comp based compositor). 

For now due to the conflict this will massively cause, I will not be able to merge this unfortunately. When implementing this later with the new UI / other changes, I will make sure to include you as a co-author if I use any code provided here. 

Although the code is most likely good and would be merged (when reviewed properly) without the previous circumstances, I would also encourage writing code your self instead of using AI for all of it, or at least in this case, making sure you understand and review the code if you didn't already. 

Again thanks for the contribution, I am sorry that we have not finished the prerequisites for this yet.

I am closing for now due to the above.

anishgoyal1108:
Thanks for the support and your advice on AI usage. It's a great application, and I hope to contribute more to it someday!

PR #186, opened 2026-06-05
Per-profile Goldberg SteamID pinning + save status/import

anishgoyal1108:
## What

Make per-profile save identities stable and manageable for Goldberg (Steam-emu) games.

### Per-profile pinned SteamID
- Each profile pins a stable, valid SteamID64 derived from its name (FNV-1a — a fixed hash, not `DefaultHasher`, since this id becomes the on-disk save-folder name and must stay constant across toolchain upgrades). The id is written to the profile's Goldberg `configs.user.ini`, so its identity — and therefore its game-save folder — is reusable across launches.
- The Profiles page shows each profile's SteamID and lets you set a specific one (e.g. to match an existing save you're importing). Pre-existing profiles get a pinned id on next launch.

### Save status + import
- New optional `Handler.save_dir` template (with `$STEAMID`), e.g. `AppData/Roaming/EldenRing/$STEAMID`. When set, the Profiles page shows a per-profile has-save / no-save indicator for the selected game and an "Import save" button that copies an existing save file into the profile's save folder.

## Testing
`cargo test` — 4 unit tests covering SteamID validity/determinism, Goldberg ini parsing, and `save_dir` resolution.

## Note
Touches `display_page_profiles`, which is also extended by the EOS PR (`feat/eos-emu-support`); if both land, that function needs a trivial merge.

### DIFF
--- a/src/app/app_pages.rs
+++ b/src/app/app_pages.rs
@@ -96,34 +96,123 @@ impl PartyApp {
 
     pub fn display_page_profiles(&mut self, ui: &mut Ui) {
         ui.heading("Profiles");
+
+        // Each profile owns a pinned Goldberg SteamID, which fixes its game-save
+        // identity (e.g. the EldenRing/<SteamID>/ER0000.co2 folder). If the
+        // currently-selected game declares where its saves live (`save_dir`), we
+        // also show per-profile save status + an import button for that game.
+        let save_ctx: Option<(String, String)> = self
+            .handlers
+            .get(self.selected_handler)
+            .filter(|h| !h.save_dir.is_empty())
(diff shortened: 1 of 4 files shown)


davidawesome02-backup:
That is an interesting note I didn't previously consider that the hashers would change, I assumed they were stable. I think most likely we should implement rust-stable-hash or a similar official hasher instead of implementing ourselves so we can use it otherwise in the project source. 

I am not sure about the way saves will be done in the future as I have considered moving to a profile oriented system in the new updates (where we just have a single full system overlay instead of a game dependent one) and that would essentially invalidate the icon...

I will need to come back to this in a little bit when I have more time to review exactly what you intend to change.

Thanks!

PR #187, opened 2026-06-05
Epic Online Services support via the Nemirtingas EOS emulator

anishgoyal1108:
## What

First-class support for games whose multiplayer runs on **Epic Online Services** (e.g. the Backrooms games) via the Nemirtingas EOS emulator, the same way `use_goldberg` supports Steam-emu games.

- **Per-handler EOS toggle + editable config** in the handler editor ("Uses Epic Online Services…"): log level, LAN broadcast discovery, loopback-only discovery, DLC unlock, overlay, language. Defaults are tuned for two instances co-op on one machine.
- **Per-instance generated config.** PartyDeck shares one overlay across instances, so a static `NemirtingasEpicEmu.json` would hand every instance the *same* EOS identity — and the emulator then filters the others' lobbies as "self", so nobody can find anyone. At launch this now **generates** a per-instance `NemirtingasEpicEmu.json` (the nested schema the current emulator reads) with a distinct `UserName`/`EpicId`/`ProductUserId`, written next to the executable. No hand-edited JSON.
- **Identity inherits the profile name**, overridable per profile on the Profiles page.
- **Auto DLL override.** When EOS is enabled, `EOSSDK-Win64-Shipping=n,b` is merged into `WINEDLLOVERRIDES` automatically (merged with any the handler already sets), so enabling the toggle is all that's needed. The user still drops the Nemirtingas `EOSSDK-Win64-Shipping.dll` into the handler overlay.
- Handlers whose overlay already ships a `NemirtingasEpicEmu.json` auto-enable on load (backwards-compat).

## Testing
`cargo test` — 7 unit tests (EOS id validity/determinism/distinctness, nested-config generation, the recursive config finder, and per-instance write-out).

## Note
Touches `display_page_profiles`, also extended by `feat/profiles-saves`; if both land, that function needs a trivial merge.


davidawesome02-backup:
I am quite interested in this, as stated on the other PR you opened, we are doing massive refactors right now, but seeing as this is not related to those (except small UI fixes) I think this could be something added without too much difficulty. I will review the code more in a few days, and may want us to change how some of the executable / sections are laid out, but this is really interesting as a extra feature. 

Thanks! Will update in a bit!

anishgoyal1108:
I think one problem with this approach is that it relies on an esoteric DLL file I found on [cs.rin.ru](https://cs.rin.ru/forum/viewtopic.php?f=29&t=105551&start=720) (a.k.a. "Nemirtingas Epic Emulator") to enable local connectivity using a single user account on Steam. I'm not sure precisely how it works on a low level, but I think it intercepts network calls made by the Epic Online Services SDK and translates them into something that other clients using the DLL can understand.

I am not sure what the ethics of this are, but I think PartyDeck's philosophy is that as long as the game is a genuine copy (*i.e.,* not a cracked copy), it should be OK, right? At least, that seems to be the main philosophy of SplitScreen.me and Nucleus. 

Important note: I didn't include the DLL in this commit for safety/privacy reasons, but you can get it directly from the link I sent in the first paragraph. If you look at the behavior from this commit, the program relies on the user to place the DLL in a very specific location within the game files. I didn't exactly write this to be user-friendly, but I hope you can build on top of it if you *do* decide to include it. I think it could be good, since in theory, this commit should enable support for any game with crossplay between Steam and Epic Games titles, not just the Backrooms-style games I mentioned earlier. According to SteamDB, based on certain file names, this is [a list of all the games that use EOS services that we could potentially port to using PartyDeck](https://steamdb.info/tech/SDK/EpicOnlineServices/).

davidawesome02-backup:
Hey, still haven't looked into it, but from my brief look, we probably cant include this in the project with the current state. I found two versions of the listed project as opensource emulators for epic services, but both are over 5 years old and are questionable. We can look into using the code or making our own version of it, but this will not be something we can easily do from the looks of it. From our current project perspective, as long as you own the game and are not breaking TOS, launching multiple copies of it on a single system would be allowed. This is known to be perfectly fine for many games like minecraft and other stand alone games. We provide a copy of Goldberg emu that allows for running multiple copies in offline mode, or with local networking as long as the games are owned and they have no further copy protection. This may be technically a grey area, but given the point of this emulator is not to pirate or bypass any TOS measure, the act of replacing one third party service (steam) with another equivalent service should be fine. Should we have a similar non-invasive replacement for 3rd party code possible from epic online services, that would require a bit more time to look into, and depending on the way such system may operate may not allow for even local multiplayer, but that is not a requirement for it to be useful for many. Would need to look into it more, I don't know yet.

Do you know if any of the Epic Games Services games can run without an internet connection - or a offline mode to allow you to start them without an internet connection? And do we have Linux native versions of these games or are most only windows versions?

From what I see briefly EpicOnlineServices for many games is just a third party thing that is not really required unless you want to have local multiplayer for many of these games. It just would mean each game would have to play independently currently.

Let me know if I am wrong @anishgoyal1108

anishgoyal1108:
You are _correct_ that the versions online are questionable and old. But the DLL I found and linked to on the cs.rin.ru forum is less than two months old and is essentially a revamped version of those projects. It is confirmed to be working on the Escape the Backrooms game and Backrooms: Escape Together. Crucially, however, it is **not** open source (at least, to my knowledge). 

As for your final question: some EOS games can run without an internet connection, and many have offline/LAN modes, like the two games I mentioned earlier that worked with this commit. However, it's up in the air whether they will work with anti-cheats or if they're not native to Windows. Though that seems to be an underlying issue with gaming on Linux in general. I was able to host a party and play the game in person on multiple monitors at my house. So, it does seem to work. We would have to evaluate whether it works for all individual games that use EOS, checking its compatibility with various compatibility layers, anti-cheat, connectivity, etc., and gauge whether they have 1-to-1 feature parity with the fully online versions. We could make a table for this, detailing the problems (if any) for those games. 

Also, I think the emulator DLL uses a different networking protocol than the Steam API. Because the games weren't being launched through Steam when I had them set up this way. So, even if the game doesn't have a dedicated "offline/LAN" mode, chances are, it can still work as long as it uses a peer-to-peer protocol for lobby creation, rather than dedicated servers.

> From what I see briefly EpicOnlineServices for many games is just a third party thing that is not really required unless you want to have local multiplayer for many of these games

You are correct about this. From what I've seen, EOS primarily enables multiplayer for games with cross-play between Steam and Epic Games. Or sometimes indie game developers use EOS purely for convenience because the SDK is easy to use for its networking features. It's pretty hacky, but it would work and add a lot of functionality to many games that no longer have dedicated multiplayer servers and have switched to using the EOS protocol for P2P networking. Some recent examples are below (credit to Claude):

> Splitgate (1047 Games) — the original Splitgate moved to peer-to-peer on August 29, 2025, as the studio needed to focus resources on rebuilding Splitgate 2. The studio said the dedicated servers had "cost us hundreds of thousands of dollars over the past couple of years," money that would be better spent keeping the dev team intact. Rather than shutting multiplayer down entirely, online play continues through P2P hosting where players themselves act as hosts. Interestingly, the Stop Killing Games movement played a significant role in shaping this decision. [PC Gamer + 3](https://www.pcgamer.com/games/fps/splitgate-fans-rejoice-peer-to-peer-multiplayer-is-coming-next-week-so-you-can-keep-on-playing-after-the-servers-shut-down/)
> CyberWar (indie, itch.io) — the original server-based version was shut down due to funding constraints, and the devs completely rebuilt the game's networking to support peer-to-peer multiplayer via NAT traversal using EOS as the subsystem. One player hosts the session, and up to 10 players can join.

davidawesome02-backup:
As for now, I would not be interested in adding anything we cant see the source code for into this project as to avoid becoming a malware distribution center. If you can find open source code for a maintained EOS emulator I can fully review it may be interesting, but until then, I don't want to cause our users problems. The api doesn't seam too complex, so it might be possible to implement a minimal working version for linux native games but that depends. From my googling, I do believe EOS implements both P2P and wider play, some of the features requiring a epic games account. I cant currently look into EOS because although I believe I can get the .h files, otherwise it would be more towards against their dev TOS to use it after downloading in such a way, given they don't publicly distribute the EOS service nor its source. If I were to implement it, it would only be from the documentation provided by them or the .so file's export symbols as to avoid legal issues. 

If you could point to some games that are (preferably) linux based, use EOS, don't strictly use online / company hosted servers / used as just a relay, and don't work offline mode SPECIFICALLY because of EOS not running that would be helpful. This would only really work for p2p games anyway given how we would have to make it work, but we would need more info. 

We just need to set what exactly we are solving with this, and what the problem is we can solve. Most games with online services are either P2P, P2P with a relay, or are hosted games. P2P should already work without any modification, hosted ones we cant do anything about, and only P2P with a well known standardized relay could really be affected here, and I am not convinced its as big of an issue currently.

Steam was a large issue because, although not drm, many games (most steam games) require it to run, and isolating this way is easiest. I think the same might be said about some EOS games because it has achievements and stuff, but I am not sure. 

Basically I just want to know the scope of the problem and the solution.

For now I think because of the above issues we should close this as a PR (just because its not in a good-for-the-project state) but please feel free to reply or send messages here. Its a really interesting idea, but its just not feasible in this way currently.

PR #188, opened 2026-06-08
Add SteamOS Holo build target for Steam Deck

jackhric:
Latest PartyDeck build couldn't be run on SteamOS due to current rolling release build CI (bleeding-edge glibc 2.43 vs SteamOS 3.7's glibc 2.41).

This PR configures and adds a CI pipeline / build image to specifically target SteamOS such that compatibility is guaranteed in the future. Creates an artifact just like other workflows in this repository.

Note: The current build-release.yml hasn't been touched; meaning automated / versioned releases won't include the Steam Deck artifacts unless explicitly uploaded to the release. This can be incorporated in a future PR.


davidawesome02-backup:
Nice, looks like it is mostly good.

Few notes if you don't mind modifying them or having me do it, the dockerfile appears to only give access to the appimage (`CMD ["bash", "-c", "export CARGO_TARGET_DIR=${CARGO_TARGET_DIR:-$(mktemp -d)} && cargo build --release -F build_gamescope -F download_deps && ./dist/build_appimage.sh"]`), maybe we should make that provide the normal binaries and / or the appimage. The AppImage can work on any (recent) OS already, so compiling on SteamOS has no real bearing on that particularly, where it is more important to build and release the raw executables that we need for special for SteamOS (In this version, executable are hidden from being easily accessed because temp dir, but I think you may be able to copy during AppImage building, but that's a bit of a mess). I think that would ideally be done by making a new directory (maybe inside of dist like how AppImage is compiled currently) to allow for extracting the partial executables or building multiple times quickly inside of the docker container instead of being a oneshot full rebuild as you have now.

Do you have a reason for disabling pipewire in this version? We don't currently make use of it, but we have some possible features planned that may use the pipewire interface exposed here and I would prefer to keep it enabled unless there is a good reason to remove it. "to stay buildable on-device" implies steamos for some reason cant download pipewire, but I am not sure if that is what you mean...

Also, I think we should make SteamOS / HOLO the one ran on release (replace the current arch build) and make said arch build be an optional github action. 
 
If I have time in the next little bit, I will post those changes if you don't do it first here so we can get this merged for the next release.
 
Also just a quick final note, I looked into the base image, and although its rarely updated (last 3 months ago), it is maintained by decky.xyz, a relatively large project, and I think they are generally trust-worthy for this currently, and being slightly out of date is acceptable for us. If we really want to be security minded, we may want to pin to specific release to avoid modification (should be noted ghcr appears to allow updating historic tags, so hash targets would have to be used). Again, that may be overkill and is not a merge blocker, but if we are trying to prevent issues we may want to pin this.


Thanks again! Happy to see you interested in the project!

jackhric:
Thanks for the review!

Agreed on the new Dockerfile. I think we should just target SteamOS binaries for Dockerfile.holo. Makes more sense that way. Working on some changes and will come back with some updates that we can discuss. 

SteamOS build replacing the current primary build sounds good too. Will retain the Arch build so it's optional.

Was having some compilation issues last night with Pipewire + the Decky baseimage, and didn't see it being used anywhere in PartyDeck yet. I'll add support back in.

And yes, we should pin to a hash. I'll add that in the PR update too.

Of course! Excited to be contributing!

jackhric:
Just finished up some work on my branch.

Fixed some issues:
- Rewrote Holo Dockerfile, added some build / run commands at top too for quick reference
- Pinned base Dockerfile image
- Pipewire build errors resolved and fixed, build flag removed
- Added build_dist.sh for use in building the project + organizing into the expected release layout (Holo Dockerfile uses this as an entrypoint), builds into release format
- Rewrote the GitHub Actions workflows. On tag, AppImage + SteamOS binaries are automatically created. Parallelized too so it's a bit quicker. They utilize the already existing Dockerfiles and scripts instead of redeclaring everything in workflows. Looks much cleaner. I can clean up the other workflows too in another pull request.

Let me know if there's anything else to fix for this PR.

jackhric:
Quick update, just fixed a naming oversight on my end for the tar content.

davidawesome02-backup:
Okay, I added some CI that runs, I am not 100% sure that works. It appears to be running right now on this branch, so I think that means we are good, but honestly I am not sure if its running twice (once for the commit and once for the PR) so we should confirm that that is correct. Also if anyone could similarly test if the executable generated here work properly, that would be helpful!

Of note, we may need to ask wunnr to confirm the workflow settings are not going to cause us problems (he has repo edit permissions) but they should be fine by default.

I moved the release build to have a common name (so we don't have one with - and one with _)
I moved the build dir to a `_generated` location as to make it obvious when running
I moved the whole build system for CI and Releases to use the main scripts.
I moved the docker file for appimage to be noted as the appimage one
I added caching for downlaoded rust objects by just adding rust home as a place inside of the generated build
I switched away from building twice as you did previously (We only really need one holoiso build then just to bundle the flatpak. The only problem I had was that holo has a older verison of `libdisplayinfo` so I just added `ln -s /usr/lib/libdisplay-info.so /usr/lib/libdisplay-info.so.2 || true` and I hope this doesn't cause problems.
I accidentally created a branch on the main partydeck project named pr-188 that I promptly removed (oops lol)
I made all the CI and test builds named "Unofficial" so we have a clear mark of "Hey this is not supported"
I fixed a spelling mistake I made before :P
 I removed the unneeded git ignore line
 
 Okay.... sorry that took so long, github actions are a pain to diagnose.

wunnr:
@davidawesome02-backup not super familiar with the workflow stuff, let me know if there are any specific settings I should change that could potentially be problematic. Otherwise LGTM

jackhric:
Just threw in my changes. Took David's changes and mine and I rewrote the workflows and dist/ completely, this is (in my eyes) the best we can get it for now. Simplifies and tidies things up a bit.

- Removed the dual Dockerfile approach and simply consolidated to one (Holo)
- Refactored and reorganized dist/
- Centralized the build action in GitHub Actions, simplified CI and version deploy, removed unnecessary workflows
- Ensured @davidawesome02-backup's primary features are intact (caching, build organizational structure, unofficial labeling, etc.)

Also @davidawesome02-backup for that libdisplayinfo error, are you using Docker to build your local AppImage development versions of PartyDeck? Haven't ran into libdisplayinfo errors while building using the image. It's all I use to compile.

davidawesome02-backup:
Hey, thanks for the update, sorry for my delay. While I do like some of the changes you made here, I unfortunately prefer some of the ones I previously added. I don't know the best way to structure this, so I will just give a general list of things I prefer from this one, and things I prefer from my changes;

In this version, you used actions, and honestly that's probably better in most cases, and I think that would be good for us to use for this, but I think it also depends a bit. For this, you also removed the builds on arch and the fedora build, I somewhat wanted to keep these as they are good reference points for our main development versions. I at least develop on arch, but using either arch or fedora as a base for the actual development I think is nice to keep around although not too important if you think its cleaner without.

I would also prefer to keep the clutter in the dist dir to a minimum, where I think here we have a few changes that are just not needed. Those are mainly: `dist/scripts/build_all.sh` being just two lines, and for most people just not needed outside of the github action or maybe the Dockerfile (most people would compile on host, and not use the compile command here, hence why the default rust dir was the users default cargo compile dir :P)

On that note, the pacman_shim.sh does absolutely nothing as far as I can see, both times being installed / registered in different ways and after all pacman operations finished? That makes no sense to me.

I would prefer to have a dockerfile for compiling the appimage separate because some people may want to make a appimage after compiling on host (to avoid clutter) while also staying the same as the host binary. 

And if possible, the reason I implemented the rust cache my self is to avoid untrusted dependencies as much as possible. While I somewhat do think we can probably trust Swatinem/rust-cache@v2, I think its a lot better to just cache simply ourselves given the minimal overhead and simplicity in the operation, but this is probably just paranoia.

Sorry if any of my comments here don't make sense, and I really don't mean to discredit your work in saying that I prefer the other changes. Thanks again, let me know what you think @jackhric

jackhric:
Hey David, no problem! We can roll back to prior changes or if you'd like, you can build on top of these. I'll explain a bit first: Where a lot of my work is coming from is trying to consolidate and standardize the development / build / release processes with what's seen in some other OSS projects and work environments.

### For GitHub Actions / CI/CD:
I use it daily at work, and it's a godsend for automating tests and builds. For example, I work on a dev branch, implement a new feature, and merge my changes into main. Upon merging to main, GitHub Actions takes it from there and builds, tests, and bundles / creates the release and sends out to clients. It's extremely useful in our environment, and was hoping to bring some of that automation to the table for PartyDeck.

### On consolidation and having just one build target / image (i.e. removing Arch and Fedora CI/CD, only one Dockerfile):
I'd like to see standardization across the build / development process such that we get a rock-solid binary every time, no matter what the developer's machine looks like. By using these build images during development (i.e. building your development changes using the image instead of your host machine by using the docker run command), you can guarantee that your binaries will run on both your machine and the Steam Deck, preventing a release event like v0.8.6 where the Steam Deck was unable to run the binary.

### On using Holo as the "single build target":
The reason I'm targeting SteamOS / Holo so heavily (and trying to make it the primary development environment with the image, removing the other build targets like Arch and Fedora) is due to the fact that they intentionally stay behind bleeding edge (fingers crossed, less potential for breakage), and they're a gaming focused distribution. They also have a massive install base, and I was interested in scaling and building awareness in the Deck community towards this project. By lowering barriers for install and getting going (i.e. making it "just work", hoping that's where [PartyDecky](https://github.com/jackhric/partydecky) can step in a bit), then this project could grow bigger fast. That would be extremely helpful for getting more people involved and building up new handlers alone.

In hindsight, I probably should have opened these changes in another PR, my changes are getting far out from just adding a SteamOS build target. This is my first time collaborating on OSS, so I'm a bit new to the process.

### As for the comments:

- `build_all.sh` can be removed. Agreed, just a helper script.
- For `pacman_shim.sh`: This allows the AppImage build to go through on Holo. `build_appimage.sh` currently runs `get-debloated-pkgs --add-mesa --add-vulkan`, an upstream script from pkgforge-dev's Anylinux-AppImages. It downloads the bleeding-edge Arch mesa/vulkan packages and bundles them into the AppImage. Now that the AppImage build runs on the Holo image (for consolidation), there's a version mismatch: those current packages declare dependencies on libgcc, libstdc++, and vulkan-mesa-implicit-layers, and the older Holo image can't resolve those names, so pacman aborts the install and the build fails. The shim is a thin wrapper for pacman that adds --assume-installed for those three names so the build goes through. I checked each one to make sure that's safe: libgcc and libstdc++ are present and are installed and owned by gcc-libs, Holo's older gcc-libs just doesn't declare those names. vulkan-mesa-implicit-layers is the odd one out and it isn't present at all, because I believe the debloated build from Decky.xyz strips it. That's a couple of optional Mesa Vulkan layers (multi-GPU device selection + a debug overlay), so skipping it doesn't affect rendering. Proton, DXVK  and native Linux games also do their own GPU selection anyway. If we were to inline it into the Dockerfile, it'd be a bit more confusing:
```
case "$1" in
    -S*|-U*)
        exec /usr/bin/pacman \
            --assume-installed libgcc \
            --assume-installed libstdc++ \
            --assume-installed vulkan-mesa-implicit-layers \
            "$@" ;;
    *)
        exec /usr/bin/pacman "$@" ;;
esac
```
I recommend the shim (or inlining into the Dockerfile) + consolidation down to Holo, but if you're looking to jump back to Arch for building and bundling the AppImage separately, then I don't see a problem either. Would solve the workaround here.

- For the Rust cache, I just saw this action was specific to caching, and reduced line count. Cool to remove as well and roll manually.

I'll be pausing all work from myself on this PR, and would consider this to be set from my side. You're free to jump in and revert, change things around, etc..

Thanks!

davidawesome02-backup:
Cool, I will most likely make some modifications where I will ping you just so you can confirm before it would be merged (it has your name attached, might as well make sure your good with it). As for the CI/CD I do think this is important, and that's something I wanted to keep and make sure it works well, that's why I added the CI run on PR or on merge :P, just so we don't have problems for official releases. On that note, I had wanted to keep Fedora and Arch as different install bases to make sure we don't selectively make it only compile on HOLO. I don't think the Fedora one has been maintained as of yet, but the Arch image at least is important for development. Implementing it in a (at least occasionally tested) build on different distros would be helpful for those wanting to participate in development or just running it on non-Arch distros. Its for that reason I think having a arch, HOLO, Fedora, and Debian (most likely) build process would be good. This would just allow us to know if any features into the future (for Arch) are broken, or if potential users on older distros may be unable to build (Debian / Fedora). I am not that strict in this, and it may be worth while to just keep arch and HOLO for now as build targets as we make such large changes to keep our target base small, but in the long run I think this is important.

And yes, I think implementing the HOLO target is quite important to make sure we don't have the same issue as 0.8.6, where I failed to properly test.

As for the pacman shim, I honestly just missed the comment saying it was explicitly for that script, that is just fine for us doing it that way.

On a completely different note, knowing how tone is difficult to give off in text as such, I mean no malice or disrespect towards your code or inclusions into the partydeck project, and am only being such a hassle for the build process update as to avoid potential problems into the future so we can have a more ideal solution. 

Thanks again! I will be making a few small changes to this version in the coming days, but otherwise looks fine to merge after I get Blahkaey to review and do my own final changes.

jackhric:
Sounds good, excited to see the changes + merge!

davidawesome02-backup:
Sorry this took so long, I think this should update be good. Fedora is just the same as before, but we dont currently use that one for almost anything. Let me know what you think about everything else. I have tested and all of the others build currently. We needed to add patchelf because that is required for do appimage builds with sharun, but otherwise the deps worked.

jackhric:
Changes look great to me. 👍

.github/actions/build-partydeck/action.yml
```diff
@@ -0,0 +1,65 @@
...
+    value: partydeck-${{ inputs.version }}-anylinux-x86_64.AppImage
+
+runs:
+  using: composite
+  steps:
+    - name: Cache pacman packages
+      uses: actions/cache@v4
+      with:
+        path: /var/cache/pacman/pkg
+        key: pacman-holo-${{ hashFiles('dist/Dockerfile') }}
```
Blahkaey: dist/Dockerfile was removed so the hashFiles() just returns "" and the key is always "pacman-holo-"

.github/actions/build-partydeck/action.yml
```diff
@@ -0,0 +1,65 @@
...
+            lcms2 hwdata seatd xcb-util-wm xcb-util-errors libdisplay-info \
+            cmake libavif libheif aom rav1e libdecor libxdamage luajit libva \
+            libxss wget curl xorg-server-xvfb fuse-overlayfs bubblewrap zsync \
+            pulseaudio pipewire-audio alsa-plugins libxkbcommon-x11 libvdpau \
+            libvdpau-va-gl sudo pkgconf openssl zip base-devel patchelf
+        pacman -S --noconfirm --overwrite '*' libpipewire
+        install -m755 dist/scripts/pacman_shim.sh /usr/local/bin/pacman  # see dist/scripts/pacman_shim.sh for more info
+
+    # Default cargo locations so rust-cache can pick up the registry + target.
+    - name: Cache cargo + build artifacts
```
Blahkaey, line 47: rust-cache strips workspace crate artifacts by default, so it will remove the gamescope build dir before saving the cache.  setting cache-workspace-crates: true should keep it and let ninja run incrementally on the next run

.github/actions/build-partydeck/action.yml
```diff
@@ -0,0 +1,65 @@
...
+      uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # 2.9.1
+      with:
+        workspaces: . -> target
+
+    - name: Build skeleton + AppImage
+      shell: bash
+      env:
```
Blahkaey, line 54: Pretty sure step level env is not visible to other steps, so the rust-cache step resolves CARGO_HOME to the default $HOME/.cargo while every crate download actually lands in workspace/.cargo. The registry never gets saved and every run re downloads all the crates.

Adding this before the rust-cache step and removing the env block from the build step would make rust-cache and the build use the same paths:

```yaml
- name: Set cargo paths
  shell: bash
  run: |
    echo "CARGO_HOME=$GITHUB_WORKSPACE/.cargo" >> "$GITHUB_ENV"
    echo "CARGO_TARGET_DIR=$GITHUB_WORKSPACE/target" >> "$GITHUB_ENV"
```

Blahkaey:
Everything else in this pr looks good. The two cache bugs are the only things I would want fixed before merge, the gamescope caching is just a nice to have that would speed up every run

davidawesome02-backup:
I will fix the cache issue by just using the string "pacman-holo" as the key, and for the other issue of the incorrect cache home, I will just remove the line as its not really needed here. Most likely we should be moving away from building gamescope here and into its own repo's build as a release eventually so we don't have cache issues between builds etc. I will add as a feature soon so we avoid the rebuilding (-F download-gamescope or the like as the default). I will push a commit with out gamescope yet as I don't think it will work well if we modify anything to do with gamescope inside of a PR. As of now, we get free compute from github for compiling gamescope, and I think it is fine for now to stay as rebuilt instead of trying to avoid the issues this causes before a proper download implementation.

Edit:
It also should be noted, I generally trust ninja continuing the build if we had a valid state in the first place, but trying to hack in using the rust cache to backup the full build dir I don't fully trust, as well concern over previous issues with ccache messups that most likely are not applicable make me not want to add artifact cache for now.

Blahkaey:
@davidawesome02-backup Okay that is algood, merge this commit if you feel everything is sorted on it now

PR #191, opened 2026-06-27
Remove dependencies

mudkipdev:
Before: 20.5 MB binary, 27.5s build time (release), 376 unique crates
After: 11.1 MB binary, 55.7s build time (release), 330 unique crates

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -448,6 +448,28 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
 
 [[package]]
+name = "aws-lc-rs"
+version = "1.17.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5ec2f1fc3ec205783a5da9a7e6c1509cc69dedf09a1949e412c1e18469326d00"
+dependencies = [
+ "aws-lc-sys",
+ "zeroize",
+]
+
+[[package]]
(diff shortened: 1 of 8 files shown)


Xyphyn (changes requested):
Delete this entire PR

davidawesome02-backup:
Hey mudkip, thanks for the commit. The main things I would want to change before this gets merged are just updating it now that it conflicts on the build_appimage script, getting confirmation that the new messages boxes work (I will test in a little bit) and most likely the removal of 
```
[profile.release]
opt-level = "s"
lto = true
codegen-units = 1
strip = true  <---
```
I know this does make the code compile into a smaller binary as you have mentioned, but unfortunately it looks like this may be at the cost of debug ability in production. Because we are still in the testing stages of this project, and implementing possibly breaking features, I am not sure if the savings on binary size is worth the striping. If we want, we can keep the lto and opt-level="s" as well as codegen-units if we want to target a smaller binary size.

Xyphyn:
You should also consider the fact that this apparently doubles the build time?

davidawesome02-backup:
Hey Xyphyn, yes that is a slight concern, but currently the largest part of our build time comes from building gamescope (planned to be moved out of this process soon), where it currently takes ~11 minutes on github actions. (seamingly ~8 of those to building gamescope vs 1-2 mins for partydeck [numbers guesstimated]) - https://github.com/partydeck/partydeck/actions/runs/28732329413

Seeing as this release build would primarily be built for real releases on github actions (we don't pay for) this may be a decent trade off of storage space vs compile time, however I do not want to cause problems for end users.

PR #198, opened 2026-07-10
added Bazzite specific instructions for new users

That1LinuxGuy:
Added Bazzite section to "Installing & Usage" portion.


That1LinuxGuy:
Bazzite Testing is not sufficient for new users. Will push docs changes correctly when I have found a better method

PR #199, opened 2026-07-11
Add controller D-pad navigation on the Instances (player-setup) page

JackTYM:
## Problem

Once a controller is added as a player on the Instances page, it can no longer navigate anything. On that page all gamepad input routes to `handle_devices_instance_menu`, which only maps buttons to add/remove/invite/launch and emits no navigation — so there's no way to set each player's **profile** (or monitor) with the controller itself. You have to reach for a mouse/keyboard.

## Change

A controller already assigned to a player now cycles **its own** player row with its **own** D-pad:

- **◀ / ▶** — cycle that player's profile (`Instance.profselection`)
- **▲ / ▼** — cycle that player's monitor (`Instance.monitor`), when the gamescope SDL backend is enabled

Because instances are keyed to devices by index, each pad only affects its own row — multiple players configure independently, with no shared focus cursor and no conflict with the existing A/B/Y/Start mapping (the D-pad was previously unused on this page). Unassigned controllers' D-pad stays a no-op.

Also adds four D-pad arrow icon assets and legend hints on the Instances page for discoverability (the monitor hint only shows when the SDL backend option is on).

## Notes

- Implemented in `handle_devices_instance_menu` using the existing `find_device_in_instance` helper; new arms deliberately fall through to the loop's `i += 1` rather than using the `continue` pattern that skips the increment.
- Modular arithmetic wraps at both ends without underflow.

## Testing

Built with `cargo build --release -F build_gamescope,download_deps` and verified on hardware: added controllers as players and confirmed each pad cycles its own profile/monitor independently, icons render, and B/X (remove), Y/[A] (invite), and Start (launch) still work.


gabriel-quina:
Can you make instances remember the last-used profile?

JackTYM:
Good idea, will add

gabriel-quina:
Hey, sorry — I got impatient and ended up implementing the profile persistence separately in #201.

I opened it as a draft because I don’t want to step on your work. I also tested #199 together with #201 locally, and both seem to work fine on my machine.

Feel free to keep working on your version. I’m happy to close mine, adapt it, or reuse whichever implementation makes more sense once yours is ready.

JackTYM:
No problem at all! I didn't get the chance to start it yet

davidawesome02-backup:
@Blahkaey Do you have any reason we shouldn't merge this as a quick thing before the next update? I think it should be fine.

Also just so yall know, we are working on a larger update to the whole UI so this will be replaced in the next large update, but should be fine for now.

Sorry I took a bit to respond, didn't get a chance to review fully until now.

wunnr:
Hi @JackTYM, i edited the UI a little bit so the dpad hints are on a separate row from the add/remove hints, just so everything fits nicely on the default window size. Otherwise everything looks good to me, thanks for contributing!

Issue #192, opened 2026-06-28
[Handler Request] Ultrakill


helphelpmeme:
Tried to write my own handler, installed the multiplayer mods with utilities
Partydeck refuses to read the json file so wondering if anyone else would have better luck

PR #200, opened 2026-07-16
Add virtual gamepads for controller reconnects

gabriel-quina:
## What this changes

Adds optional virtual gamepads backed by `uinput`.

When enabled, Partydeck creates a virtual controller for each physical controller and forwards input events to it. Game instances receive the virtual devices instead of the physical ones.

The main goal is to keep controller assignments stable when a physical controller disconnects and reconnects.

## Implementation notes

- Adds a `use_virtual_gamepads` option, disabled by default.
- Creates and manages virtual controllers through `uinput`.
- Forwards events from physical controllers to their virtual counterparts.
- Keeps the virtual device available while the physical controller is disconnected.
- Reconnects the physical controller to the same virtual device when possible.
- Exposes only the assigned virtual controllers inside each game instance.
- Filters Partydeck's own virtual devices from normal controller detection.

## Testing

Tested locally with multiple instances and controllers.

I checked:

- disconnecting and reconnecting controllers;
- keeping controller assignments after reconnecting;
- isolating controllers between instances;
- launching games with the option enabled and disabled.

This currently works on my setup, but testing with other controller models and Linux environments would be useful.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2704,6 +2704,7 @@ dependencies = [
  "evdev",
  "fastrand",
  "image",
+ "nix 0.29.0",
  "pathsearch",
  "reqwest",
  "rfd",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,7 @@ zbus = { version = "5.5.0", features = ["async-io", "blocking-api"], default-fea
 steamlocate = "2.0.1"
 zip = { version = "5.1.1", features = ["deflate", "deflate64"], default-features = false }
 pathsearch = "0.2.0"
+nix = { version = "0.29", features = ["poll"] }
 
 [build-dependencies]
 reqwest = { version = "0.12.15", features = ["blocking", "json"], optional = true }
--- a/src/app/app_pages.rs
+++ b/src/app/app_pages.rs
@@ -639,6 +639,10 @@ impl PartyApp {
             &mut self.options.kbm_support,
             "Enable keyboard and mouse support through custom Gamescope",
         );
+        let virtual_gamepads_check = ui.checkbox(
+            &mut self.options.use_virtual_gamepads,
+            "Preserve controller bindings on reconnect (Experimental)",
+        );
         let gamescope_force_grab_cursor_check = ui.checkbox(
             &mut self.options.gamescope_force_grab_cursor,
             "Force grab cursor for Gamescope",
@@ -652,6 +656,9 @@ impl PartyApp {
         }
         if kbm_support_check.hovered() {
             self.infotext = "Runs a custom Gamescope build with support for holding keyboards and mice. If you want to use your own Gamescope installation, uncheck this.".to_string();
+        }
+        if virtual_gamepads_check.hovered() {
+            self.infotext = "Routes assigned gamepads through persistent virtual controllers, game instances keep their controller after reconnects. Gamepads only.".to_string();
         }
         if gamescope_force_grab_cursor_check.hovered() {
             self.infotext = "Sets the \"--force-grab-cursor\" flag in Gamescope. This keeps the cursor within the Gamescope window. If unsure, leave this unchecked.".to_string();
--- a/src/app/config.rs
+++ b/src/app/config.rs
@@ -31,6 +31,8 @@ pub struct PartyConfig {
     #[serde(default = "default_true")]
     pub kbm_support: bool,
     #[serde(default)]
+    pub use_virtual_gamepads: bool,
+    #[serde(default)]
     pub proton_version: String,
     #[serde(default = "default_true")]
     pub proton_separate_pfxs: bool,
@@ -58,6 +60,7 @@ impl Default for PartyConfig {
             gamescope_sdl_backend: true,
             gamescope_force_grab_cursor: false,
             kbm_support: true,
+            use_virtual_gamepads: false,
             proton_version: "".to_string(),
             proton_separate_pfxs: true,
             proton_wow64: true,
--- a/src/input.rs
+++ b/src/input.rs
@@ -142,6 +142,9 @@ impl InputDevice {
 pub fn scan_input_devices(filter: &PadFilterType) -> Vec<InputDevice> {
     let mut pads: Vec<InputDevice> = Vec::new();
     for dev in evdev::enumerate() {
+        if crate::virtual_gamepad::is_partydeck_virtual_device(&dev.1) {
+            continue;
+        }
         let enabled = match filter {
             PadFilterType::All => true,
             PadFilterType::NoSteamInput => dev.1.input_id().vendor() != 0x28de,
--- a/src/launch.rs
+++ b/src/launch.rs
@@ -1,13 +1,14 @@
 use std::path::{Path, PathBuf};
 use std::process::Command;
 
-use crate::app::{PartyConfig, PadFilterType};
+use crate::app::{PadFilterType, PartyConfig};
 use crate::handler::*;
 use crate::input::*;
 use crate::instance::*;
 use crate::paths::*;
 use crate::profiles::{create_profile, create_profile_gamesave};
 use crate::util::*;
+use crate::virtual_gamepad::VirtualGamepadSession;
 
 pub fn setup_profiles(
     h: &Handler,
@@ -36,7 +37,18 @@ pub fn launch_game(
     instances: &Vec<Instance>,
     cfg: &PartyConfig,
 ) -> Result<(), Box<dyn std::error::Error>> {
-    let new_cmds = launch_cmds(h, input_devices, instances, cfg)?;
+    let virtual_gamepad_session = VirtualGamepadSession::start_if_enabled(
+        cfg.use_virtual_gamepads,
+        input_devices,
+        instances,
+    )?;
+    let new_cmds = launch_cmds(
+        h,
+        input_devices,
+        instances,
+        cfg,
+        virtual_gamepad_session.as_ref(),
+    )?;
     print_launch_cmds(&new_cmds);
 
     if cfg.enable_kwin_script {
@@ -45,7 +57,8 @@ pub fn launch_game(
             false => "splitscreen_kwin.js",
         };
 
-        kwin_dbus_start_script(PATH_RES.join(script)).map_err(|e| format!("Failed to start KWin script: {}", e))?;
+        kwin_dbus_start_script(PATH_RES.join(script))
+            .map_err(|e| format!("Failed to start KWin script: {}", e))?;
     }
 
     let sleep_time = match h.pause_between_starts {
@@ -58,7 +71,11 @@ pub fn launch_game(
     let mut i = 0;
     for mut cmd in new_cmds {
         let handle = cmd.spawn().map_err(|e| {
-            format!("Failed to start '{}': {}", cmd.get_program().to_string_lossy(), e)
+            format!(
+                "Failed to start '{}': {}",
+                cmd.get_program().to_string_lossy(),
+                e
+            )
         })?;
         handles.push(handle);
 
@@ -80,6 +97,7 @@ pub fn launch_cmds(
     input_devices: &[DeviceInfo],
     instances: &Vec<Instance>,
     cfg: &PartyConfig,
+    virtual_gamepads: Option<&VirtualGamepadSession>,
 ) -> Result<Vec<std::process::Command>, Box<dyn std::error::Error>> {
     let win = h.win();
     let exec = Path::new(&h.exec);
@@ -90,7 +108,9 @@ pub fn launch_cmds(
     };
 
     if cfg.kbm_support && !gamescope.exists() {
-        return Err("gamescope-kbm is missing. Please reinstall partydeck or disable KBM support.".into());
+        return Err(
+            "gamescope-kbm is missing. Please reinstall partydeck or disable KBM support.".into(),
+        );
     }
 
     if !cfg.kbm_support && pathsearch::find_executable_in_path("gamescope").is_none() {
@@ -103,10 +123,16 @@ pub fn launch_cmds(
                 .join("steam/steamapps/common/SteamLinuxRuntime_soldier")
                 .exists())
         || (runtime == "sniper"
-            && !PATH_STEAM.join("steam/steamapps/common/SteamLinuxRuntime_sniper").exists()
-            && !PATH_STEAM.join("steam/steamapps/common/SteamLinuxRuntime_sniper-arm64").exists())
+            && !PATH_STEAM
+                .join("steam/steamapps/common/SteamLinuxRuntime_sniper")
+                .exists()
+            && !PATH_STEAM
+                .join("steam/steamapps/common/SteamLinuxRuntime_sniper-arm64")
+                .exists())
         || (runtime == "steamrt4"
-            && !PATH_STEAM.join("steam/steamapps/common/SteamLinuxRuntime_4").exists())
+            && !PATH_STEAM
+                .join("steam/steamapps/common/SteamLinuxRuntime_4")
+                .exists())
     {
         return Err(format!("Steam Runtime {runtime} not found! Runtime must be installed on the same drive that the Steam client is installed on.").into());
     }
@@ -116,11 +142,12 @@ pub fn launch_cmds(
         .collect();
 
     for (i, instance) in instances.iter().enumerate() {
-        let gamedir = if h.is_saved_handler() && !cfg.disable_mount_gamedirs && cfg.profile_unique_dirs {
-            PATH_PARTY.join("tmp").join(format!("game-{}", i))
-        } else {
-            PathBuf::from(h.get_game_rootpath()?)
-        };
+        let gamedir =
+            if h.is_saved_handler() && !cfg.disable_mount_gamedirs && cfg.profile_unique_dirs {
+                PATH_PARTY.join("tmp").join(format!("game-{}", i))
+            } else {
+                PathBuf::from(h.get_game_rootpath()?)
+            };
 
         if !gamedir.join(exec).exists() {
             return Err(format!("Executable not found: {}", gamedir.join(exec).display()).into());
@@ -177,7 +204,10 @@ pub fn launch_cmds(
             cmd.env("SDL_GAMECONTROLLER_ALLOW_STEAM_VIRTUAL_GAMEPAD", "1");
         }
         if cfg.pad_filter_type == PadFilterType::OnlySteamInput {
-            cmd.env("SDL_GAMECONTROLLER_IGNORE_DEVICES", SDL_GAMECONTROLLER_IGNORE_DEVICES);
+            cmd.env(
+                "SDL_GAMECONTROLLER_IGNORE_DEVICES",
+                SDL_GAMECONTROLLER_IGNORE_DEVICES,
+            );
         }
         if !h.env.is_empty() {
             for env_var in h.env.split_whitespace() {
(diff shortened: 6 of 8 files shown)


Blahkaey:
I think @davidawesome02-backup found there is issues when using uinput, you can read what he said in pr #176. I have experienced the issue which you are trying to solve and do think there needs to be a solution. Related to this i have found that something on my install of cachy-os has recently changed, and when a controller disconnects/reconnects, it seems to keep the same device path and successfully reconnects to the games instance, looking into what made this change could be beneficial aswell.

gabriel-quina:
I'm on cachyOs too.  Gonna upgrade and test
For now this implementation is working without any issues. Only using to play don't starve together though

davidawesome02-backup:
Hey @gabriel-quina! This is nice to see, and I am honestly interested in this problem, it has been noted as a real issue we need to solve. I do see some possible problems however with the way you are doing this that may make it less than ideal for our project although a possible solution; Using UINPUT allows for creating virtual devices, but unfortunately I don't believe it is able to *fully* emulate a device, instead allowing access to the most common inputs and data about a device. This *Should* be fine in most scenarios, but the controllers that do implement extra features with this would be unable to use them (I honestly cant find the example I want to use, but there was some controller that required a custom driver to properly emulate on linux somewhere). As a more direct approach, I was soon going to attempt to work on a `unshare` / bwrap replacement inside of partydeck to allow for binding and unbinding without restart directly on that layer. 

As another possible issue, requiring uinput (requires editing the system setting, or having steam installed to do so) is not really ideal if we want to allow non-steamos / systems without steam this becomes a bit of a hassle as a strict requirement. 

I do plan to use uinput in this project in the near future to allow for remote game streaming / inputs, but that would be an optional thing on top of the base that works without modifications to the standard linux setup. 

Also on a more personal code standpoint for this project, I would ideally like to avoid busy looping as much as possible as it just wastes CPU time when no new events come in, as well as delaying events minimally (use select / poll / epoll to get direct kernel events when they come in). You should ideally in something like this make sure to bind to rumble as well as all inputs provided by the selected controller (not just the default ones).

Also, this code segment: `pub const VIRTUAL_GAMEPAD_NAME_PREFIX: &str = "PartyDeck Internal Virtual Gamepad";` will appear to users if we actually modify the name in certain game's UIs. This is not great design especially as we have multiple controllers trying to find "what is what" may be hard, so a shorter prefix or suffix may be better

Let me know what you think as this is just my opinion, and if there is any justification to using uinput otherwise it may be something we want to take the aforementioned hits to the project if we have a real end user experience improvement we can point to otherwise. Maybe this could be an option if all else fails in addition to the remote streaming if we cant get custom `unshare` working by then.

gabriel-quina:
Thanks @davidawesome02-backup for the feedback. I agree with the issues you pointed out.

I’m not really attached to the uinput approach. I ran into this problem on my setup and tried to solve it with what I knew, and it worked well enough for my use case.

I think I misunderstood part of @Blahkaey’s comment at first. Thinking back, reconnection already worked on my CachyOS setup before this implementation, probably because the controller kept the same device path. The problem was that after reconnecting, the same controller would control both sessions instead of only the one it was originally assigned to.

So the main benefit of this approach for me was keeping each controller isolated and assigned to its original session after disconnecting and reconnecting.

Also, uinput is currently optional and disabled by default, but I understand why it may not be a good base solution for the project, especially because of the setup requirements and missing controller features.

Game streaming is also something I’m interested in. I’ve used Partydeck with Steam Link before and it worked, although I remember having to use Steam Input. I’d also like to test it with Moonlight/Sunshine, but I haven’t had the chance yet.

For now I’ll probably keep using this version locally, and I may still fix some of the problems you mentioned, especially the polling and the virtual device naming.

I’d be happy to look into the mount namespace approach you mentioned. Is there any existing issue, branch, or code I could use as a starting point?

davidawesome02-backup:
One of the main problems is that the current approach works well but because of how we are masking inputs into bwap, any new controllers or those with new ids in /dev/input/XXX will be then accessible to ALL instances, it is only those specifically masked out at the start that are currently hidden. We should invert this logic, but I believe this is best done via a little complex setup of: `dev mount /, tmpfs mount /dev/input, dev mount /dev/input/XXX` instead of our current `dev mount /, dev mount /dev/input/ALLOTHERS /dev/null`. 

Also, not sure how much you know about this, but we (I) also plan to eventually make overlayfs a global setting as to remove the "current game dir" and strange settings, so things are expected to change in the long term. (Quick summery even though this is most likely not anywhere needed information: Steamos doesn't support overlayfs due to a outdated kernel that doesn't support casefolding in overlayfs, and thus relies on fuse-overlayfs being dev mounted on only the directory we want to target. In the long term, we either want raw overlayfs for performance gains, or to implement our own fuse implementation to allow us to actions like faking the goldberg dll in certain places etc. This is just future plans however. )

As for game streaming, I believe moonlight is currently a solution, however my ideal solution, and one very possible in near updates, is based around using the frames we get from gamescope ourselves, and creating a webrtc stream to a website (cloudflare pages for free) to allow for easily accessible connections while (for example) at a friends house with this setup. It would be as simple as typing in a 6 digit code into a website on your phone or laptop, and getting access to your own screen, controller, keyboard and mouse, or just access to a remote controller on your phone if you don't have enough.

Honestly I wrote my earlier message as I was running out of the house, so didn't have time to rewrite after I saw you had that as a option and not as a required pass though, so sorry for missing that :P.

We don't currently have any version of the bwrap replacement code I would want as I have not spent the time to create one yet, however I do have the following general design goals and information that could help.

### Bwrap replacement ramble:

Honestly if you are willing to mess around with it, I would just say try to make a basic rust version of `bwrap` from the source code here: https://github.com/containers/bubblewrap (GPL v2, so we can do this) as a rust crate. The main things we need are to use `unshare` as documented [here](https://man7.org/linux/man-pages/man1/unshare.1.html), [here](https://docs.kernel.org/userspace-api/unshare.html) and [here](https://man7.org/linux/man-pages/man2/unshare.2.html) as well as [here](https://www.ibm.com/docs/en/zos/3.1.0?topic=lf-unshare-isolate-process-from-one-more-its-namespaces)

Okay, admittedly that last one is a random IBM page on unshare, but lets just pretend its helpful in anyway.

Anyway, if take a look at the existing crates, most of them are just not quite what we want as far as I can tell: https://crates.io/crates/landlock (overly complex, no custom mounting as far as I can tell, but more direct IO blocking via newish kernel code), https://crates.io/crates/unshare (
```
The following is out of scope:

    mounting file systems
    setting up network
    in-container and out of container supervision
    handing child signals
```
We require setting up file systems, and handling child signals somewhat, so unfortunately not a perfect solution, although maybe helpful?) 

Also this random project as a front end for a few things maybe has some useful code: https://crates.io/crates/procjail

The main thing I would want would be (not sure if its possible because of how "threads" exist in both rust and unshare's API, but) to have a simple object with a thread pointer as well as [a simple thread request](https://doc.rust-lang.org/std/sync/mpsc/fn.channel.html) that could pass over things like a `std::...::Command()` or the like, and return back a complex type like `std::...::Child`. The API I imagine is something like the following:

Assume all paths are most likely PathBuf or something logical
```rust
let c = API::Container::new()
c.bind("/",API::mount::dev("/",API::readonly::RW)) // Tempfs, dev bind, unmount, etc that we use right now.
c.bind_data("/data.txt", bytes_object)
c.start(_: std::Command) -> std::Child
```
(we don't currently do PID isolation nor network isolation, those may be something we want in the far future, but for now almost no games should need those in place, so omission should be fine)

After spawning a child, you will have to follow more of the bwrap source code, and implement the `unshare` syscall, the UID table writing, and process binding.

The main thing to note here is that the bind API can be called AFTER the process has been started, allowing for us to rebind, unbind, and do all the other fancy stuff we cant do with bwrap.

I honestly gave this to a bad LLM, and got some output but nothing usable to this point, and mostly following the lead of having to fork / spawn a new process to handle `unshare` semantics, however it looks like this may be avoidable and simplifies a lot of the IPC (std::Child cant be recreated on another process, but can be sent over threads via Arc mutex)

# -
Sorry this is so long, and honestly just a jumble of plans / ideas I have had as improvements on the project, if you don't have the time to work on this please don't feel pressured into making something that is just a possible good idea. We are doing a lot right now (even if not released yet), and I would hate to see us change direction again and again. Feel free to let me know what you think about any of this if you have time to read this textbook of a response.

Again however, uinput may be something useful in the future, even implementing this if we find it best, I just believe we might be able to solve two issues at once by replacing bwrap despite its added complexity.

Again, thanks for the PR even if you cant work on these more complex features, we really do appreciate the support, I just have been picky in how we do things for what I would ideally want. I would defer to Blahkaey's and Wunnr's opinions on a topic if they however believe me to be inaccurate in my assessment of any given idea.

gabriel-quina:
Thanks for the detailed explanation. Building on the approach you described, I’m thinking about one container context per running instance, each starting with an empty "/dev/input", exposing stable controller slots, and keeping a mapping like "instance → slot → physical device identity". On reconnect, PartyDeck could resolve the new host path and rebind only the correct slot in the correct instance.

I’ll go through the references you shared and look more closely at the broader container direction. I understand why it is useful to keep future requirements in mind, although the web client sounds like another beast 😅

While that is being explored, would you be open to a clearly documented temporary solution for the current hotplug issue? It is quite disruptive in normal use: having to close multiple games and redo the controller bindings after a reconnect has nearly made me stop using PartyDeck regularly, especially when playing with my wife and trying to avoid interrupting the session repeatedly.

One other thing I’m wondering about: would this approach also cover games that rely exclusively on the Steam Input API?

davidawesome02-backup:
Okay, so for the input rebinding, I would almost definitely take the has I derive (in testing code) [here](https://github.com/davidawesome02-backup/partydeck/blob/new-layout-tests/src/input.rs): 
```
pub fn hash(&self) -> DeviceHash {
        *self.hash.get_or_init(|| {
            let mut hasher = std::hash::DefaultHasher::new();
            (&(self.dev.unique_name(),self.dev.input_id(),self.dev.name())).hash(&mut hasher);
            hasher.finish()
        })
    }
```
giving us a value that should be constant between reconnects. This was added already to my test code because I wanted to allow for devices to be connected or reconnected without having to "rescan" as we do right now and reset everything, as well as allowing for saving controllers to a specific profile for later use. As this is derived from both the "input id" (described using the steam id here) as well as the name, we should get something unique for both steam and physical device for each physical device allowing for it to work without any real issue.

As such in the new UI, devices for instances are only stored using their hash, where resolution for connection is done via a quick map over the currently known input devices. It adds a few extra ms, but seeing as we have it in a oncelock, its not like its recalculating the hash every time, just a quick map over a max of 20 maximally reasonable devices.

The web client will be an separate problem, yes, but I do think feasible. The code for sending video as well as cloudflare backend as a example is already done and public on my github account, although this will be massively changed. I am in the moonlight discord server where many have tried this before, and we can get low delay with webRTC via the media API if we try (in my case, using libdatachannel and a stream that goes at EXACTLY 60 fps every time), but if we have any frame delay, the browser's jitter buffer will increase, and slowly add more and more delay to the playback. Many use webRTC unreliable data channels and the browsers media api to resolve this issue by injecting the required compressed blank frames to "skip" ahead if we miss anything. This is better, but adds a lot of complexity I have not done yet. Otherwise, I have some friends who really like front end design who can help make the UI work, because my UI design skills are terrible.

And sure, honestly I would somewhat be open to if you wanted to try to get this in as a sample fix before partydeck UI and backend reworks we are doing, but I am not sure about how we would do that in the current ether that is our development cycle. Reminds me we have to make a new release soon after I do a few quick final checks (merge #196 (with mods) and #192 (after checks) ), so maybe we can work this into that if we want to tide us over..?

Let me know what you think, thanks for the input.

gabriel-quina:
Yeah, I think it’s worth trying to get this in as a temporary fix.

I’ll keep working on the uinput implementation and try to address the issues you mentioned. I’ll also try using the device hash for reconnects and see how well it works with different controllers and connection methods.

I have two GameSir T4N Lite controllers of the same model, a wired Microsoft Xbox controller for PC, a generic PS2 controller, and a Logitech wheel with force feedback, so I can test identical devices, USB, Bluetooth, wireless, reconnects, and force feedback.

Meanwhile, I’ll also look into the other approaches you mentioned and use them as a starting point.

davidawesome02-backup:
@gabriel-quina 

> I have two GameSir T4N Lite controllers of the same model, a wired Microsoft Xbox controller for PC, a generic PS2 controller, and a Logitech wheel with force feedback, so I can test identical devices, USB, Bluetooth, wireless, reconnects, and force feedback.

Honestly if you don't mind, please do! I believe that it most likely wont work with the exact model now that I read over it again due to having no real exact refrence. Googling for a moment, it looks like we may be able to get a serial number or other from parsing the device directly or maybe from something else from evdev. I will look at evdev structs in a moment, but otherwise I only have a set of nintendo switch controllers that I own, so nothing much I can test conclusively my self. If you wouldnt mind trying to get unique identifers to add to this hash that are preferably stable across reboots, please let me know :).

We can for now add a fallback to hash, /dev/eventX path. Then if that hash is no longer at that path, search again for it, etc. But let me know what you think!

gabriel-quina:
I did a quick test with two identical controllers to see how Linux identifies them.

With the 2.4 GHz dongles, both devices exposed the same values:

```text
name: Generic X-Box pad
input_id: USB / 3537:1040 / 0100
unique_name: empty
serial_short: empty
```

The only visible difference was the USB path, since each dongle was connected to a different hub port.

I also connected both controllers directly by cable, one at a time, using the same cable and USB port. Both exposed:

```text
name: Zikway HID gamepad
product: 3537:1041
unique_name: empty
serial: Zikway_HID_gamepad
```

The differences I saw were only dynamic kernel enumeration values such as `input40` versus `input44`.

So, at least for these controllers, hashing:

```rust
(unique_name, input_id, name)
```

would likely produce the same value for both units.

Maybe this hash should be treated as a best-effort device fingerprint rather than a guaranteed unique controller ID. It may also be worth detecting duplicate fingerprints and handling them as ambiguous, possibly using the USB path or asking the user to press a button as a fallback.

### Bluetooth follow-up

I also tested both controllers over Bluetooth.

In this mode, Linux exposed a different `UNIQ` value for each controller:

```text
Controller A: 5D:6C:A4:58:BF:86
Controller B: 0B:2D:FF:61:EB:69
```

The values matched the Bluetooth addresses reported by `bluetoothctl`, so they appear to be usable per-controller identifiers while connected over Bluetooth.

One detail is that the controller identifies itself using Sony/PS4-compatible values:

```text
name: Wireless Controller
input_id: 0005:054c:09cc:8100
```

Linux also created separate endpoints named `Motion Sensors` and `Touchpad`. However, this controller is physically an Xbox 360-style controller and does not actually have a touchpad or motion sensors. This suggests that the Bluetooth firmware may be emulating a DualShock 4-compatible device profile, so the reported capabilities should not necessarily be assumed to match the real hardware.

The behavior observed so far seems transport-dependent:

```text
Bluetooth: controllers can be distinguished by UNIQ
USB cable: no individual UNIQ or serial was exposed
2.4 GHz: only the receiver/USB path was distinguishable in my tests
```

There still does not appear to be a common identifier that automatically links the same physical controller across USB, Bluetooth, and 2.4 GHz modes.

davidawesome02-backup:
@gabriel-quina 

Without a unique number, I guess we will have to do our own handling... What do you think about a mapping like the following (untested):

```rust
impl InputDevice {
    ...
    fn deviceID(self) -> DeviceID {
        DeviceID{ hash: self.hash(), path: self.path() }
    }
    ...
}


#[derive(Equals)]
struct DeviceID {pub hash: DeviceHash, pub path: String}

struct DeviceSearcher {
    devices_left: std::vec<InputDevice>
}

// Constructed once to search over devices in a stateless way, extracting the first matching from the list.
// Cant be used more than once
impl DeviceSearcher {
    fn new(devices: std::vec<InputDevice>) -> Self {
        Self {
            devices_left: devices
        }
    }

    fn find(self, id: DeviceID) -> Option<InputDevice> {
        // Fallback if the exact path isnt found with that hash.
        let mut hash_match_first = None;
        // Pulled out here because I am writing without being in the codebase and cant syntax check borrowing to remove_mut while inside the for loop.
        let mut id_match_first = None; 
        for (idx, device) in &self.devices_left.itter().enumerate() {
            if hash_match_first.is_none() && device.hash() == id.hash {hash_match_first = Some(idx);}
            if device.deviceID == id {id_match_first = idx; break;}
        }

        if let Some(id_match_first) = id_match_first {
            Some(self.devices_left.swap_remove(id_match_first))
        } else if let Some(hash_match_first) = hash_match_first {
            Some(self.devices_left.swap_remove(hash_match_first))
        } else {
            None
        }
    }
}
```

gabriel-quina:
@davidawesome02-backup 

Since the event path may remain stable on some systems but be recreated on others, could we also store `physical_path()` and use it before the hash-only fallback?

Something like:

```rust
struct DeviceID {
    hash: DeviceHash,
    event_path: String,
    physical_path: Option<String>,
}
```

And then match in this order:

1. Same hash and same event path.
2. Same hash and same physical path.
3. First matching hash as the final fallback.

That should preserve the simple case on systems like my current CachyOS setup, while still handling cases where the device comes back with a different `eventX` but the dongle remains on the same USB port.

I am not sure whether `physical_path()` has any stability issues I am overlooking, though. What do you think?

davidawesome02-backup:
Sure, I think that works if we want to try that, its just slightly longer to store as a multiple section object for compare, but I dont mind too much. This should be good as a more accurate location, and then after that fallback to others.

I out of time for right now, but will respond more later. Thanks!

Also let me know what you think of the bwrap replacement stuff if you have time. Will respond more tonight.

davidawesome02-backup:
Hey, still working on the new design stuff, and taking far too long. Just wanted to update you and tell you I had a new controller disconnect / reconnect scheme that should be fully serializable, while supporting many devices with the same hash, and allowing post-launch reconnects. I am still working on launch code, so this current one just has the input handler if you wanted to try to take a look.

https://github.com/davidawesome02-backup/partydeck/tree/new-layout-launch

turdferguson5000:
I ran into the same problem and ended up building almost exactly this, so a couple of things from my setup that might be useful.

@davidawesome02-backup is right about the masking issue, and I think it's worth separating from the virtual gamepad question, because it bites on its own. I hit it with four Xbox 360 pads on a wireless receiver. They power themselves off when idle, and waking one destroys the kernel input device and creates a new one. Kernel log from a four player session, launched at 20:36:

```
20:35:55  event24 = input257   event29 = input259   event30 = input260
20:46:49  event29 -> input264
20:46:50  event30 -> input265
```

The two players whose pads power cycled lost input completely. The one whose pad never slept was fine. Same node names and same minor numbers, different devices.

The trap for the uinput approach specifically: node numbers get recycled. When a pad powers off it frees its `eventN`, and the next uinput device created takes that number. In my case player 1's virtual pad landed on the `event29` that player 3's controller had just vacated, and player 3's device path then pointed at player 1's virtual pad. I ended up with a device named `PD Pad 3 (PD Pad 1 (Xbox 360 Wireless Receiver))`, one player's stick driving another player's game, and the real pad 3 driving nothing.

What fixed it for me was keying on the receiver slot, the USB interface in sysfs, rather than the device path. Worth knowing that with four pads on one receiver, name, vendor, product and phys are all identical and `uniq` is empty, so the interface is genuinely the only thing that distinguishes them.

I can also confirm the rumble limitation. Force feedback does not survive a plain uinput mirror, you would need `UI_FF_UPLOAD` handled and relayed back to the real device.

Separately I've opened a small PR for just the masking part, since that one is fixable without any virtual device at all.

On provenance, since it came up on #176: this was written with AI assistance and directed and tested by me on the hardware above. Flagging it rather than having you find out.

PR #201, opened 2026-07-16
Remember last selected profiles for instances

gabriel-quina:
## What this changes

Remembers the last selected profile for each instance.

The selections are stored in a separate JSON file and restored when instances are created again.

## Implementation notes

- Saves profile names instead of profile indexes.
- Restores the saved profile for each instance position.
- Falls back to Guest if a saved profile no longer exists.
- Stores the state outside the main config file.

The file is saved at:

`~/.local/share/partydeck/last_instance_profiles.json`

## Testing

Tested locally by changing profiles across multiple instances, restarting Partydeck, and recreating the instances.

I also tested it locally together with #199. Both changes appear to work correctly together on my machine.

There is related profile work in progress, so this implementation may need adjustments depending on what gets merged.

### DIFF
--- a/src/app/app.rs
+++ b/src/app/app.rs
@@ -1,11 +1,12 @@
-use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::thread::sleep;
 
 use super::config::*;
 use crate::handler::*;
 use crate::input::*;
 use crate::instance::*;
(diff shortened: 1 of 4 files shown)


Blahkaey:
Hi thanks for the pr, unfortunately david and i are in the middle of some large changes regarding how the games are displayed and how src/app/* is structured. One piece of these changes is a rebuild of the ui for adding instances, the layouts of game instances can be changed before launching and layout/profiles is stored persistently. 

Hopefully within the next week our changes will be pushed to the main repo.

Cheers

PR #209, opened 2026-08-17
Show each instance only its own input devices instead of masking others

turdferguson5000:
Follows on from something @davidawesome02-backup described in #200:

> because of how we are masking inputs into bwrap, any new controllers or those with new ids in /dev/input/XXX will be then accessible to ALL instances

This is a fix for that specific problem, with no virtual devices involved.

**The bug.** Masking can only act on devices that exist when bwrap starts. Anything that shows up afterwards was never masked in any instance, so every instance can read it. With wireless pads this happens constantly, because they power off when idle and waking one destroys the kernel device and creates a new one, usually on a different `eventN`. The player wakes their controller and starts driving all four games.

**Why it can't be patched in place.** There is no path to bind `/dev/null` over until the device exists, so no amount of extra masking closes it. The list has to be inverted.

**What changed.** `/dev/input` becomes a tmpfs and the devices that should be visible get bound back in. The rules for what is visible are deliberately identical: a device is bound unless it is disabled, or it is a gamepad belonging to another player. `jsN` nodes are bound next to their evdev node so anything using the legacy joystick API still finds them. hidraw sits outside `/dev/input` and is masked exactly as before.

**Testing.** Running daily on my setup, four players, Xbox 360 wireless receiver, three monitors, Orcs Must Die 3 and Core Keeper. Before this, a pad that woke up mid session controlled every instance. After it, it controls none until relaunch, which is not perfect but is a much better failure. I also checked with SDL inside a replica sandbox that a whitelisted pad still enumerates, both with and without udev reachable.

**Not tested:** SteamOS or Steam Deck, PS controllers or the hidraw path, and keyboard/mouse instances beyond confirming they still get their devices.

**Provenance,** since it came up on #176: written with AI assistance, directed and tested by me. I'd rather say so than have you spot it. Happy to walk through any of the reasoning, and happy for it to be taken as a bug report with a suggested patch if you'd rather write it yourselves.

### DIFF
--- a/src/launch.rs
+++ b/src/launch.rs
@@ -9,6 +9,21 @@ use crate::paths::*;
 use crate::profiles::{create_profile, create_profile_gamesave};
 use crate::util::*;
 
+/// Legacy /dev/input/jsN nodes belonging to the same physical device as an evdev node.
+fn js_siblings(evdev_path: &str) -> Vec<String> {
+    let Some(name) = Path::new(evdev_path).file_name().and_then(|s| s.to_str()) else {
+        return Vec::new();
+    };
+    std::fs::read_dir(format!("/sys/class/input/{name}/device"))
+        .into_iter()
+        .flatten()
+        .flatten()
+        .filter_map(|e| e.file_name().to_str().map(str::to_string))
+        .filter(|n| n.starts_with("js"))
+        .map(|n| format!("/dev/input/{n}"))
+        .collect()
+}
+
 pub fn setup_profiles(
     h: &Handler,
     instances: &Vec<Instance>,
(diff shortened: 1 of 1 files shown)


davidawesome02-backup:
Hi @turdferguson5000 this is an intresting change, and could be useful for our device isolation, however in the current version I am not sure if we want to make this happen yet via this process (Could be useful however). I am currently working on a full rewrite of the input system to allow for rebinding as well as having this work properly. If you have any deeper knowledge or ideas for this, please let me know here: https://matrix.to/#/!oioKjUlkaDWGuWogsV:matrix.org/$R1JYF_naVDvpVuKtVvcj8L8e5cfuNh4jvkXy4-rjZzw?via=matrix.org&via=matrix.vulkanlan.at&via=tchncs.de (https://matrix.to/#/#partydeck-dev:matrix.org). I am currently trying to make it so we can do rebinding with a seprate unshare program, but am open to comments. I currently have my tests around this same design of the tmpfs with binding as you got from my previous posts. Happy to get some input, thanks!

Will look more at the other posts in a bit, writing this in a rush :P

turdferguson5000:
That all makes sense, and no rush on this one. If the input rewrite covers it, close this and I will not be offended. It started as a bug report that happened to come with a patch.

Good to hear the tmpfs plus bind design matches what you are already testing. I did not know that when I wrote it, so treat this as a second data point rather than a proposal.

Since you are building rebinding on top of a separate unshare program, here are the things that cost me the most time. They all bite harder with virtual devices than with real ones, so I think they are relevant to the rewrite specifically.

**1. Node numbers get recycled, and virtual devices are what land in the freed slot.**

When a pad powers off, its `eventN` is freed. The next uinput device created gets that number. In my case a pad was restarted mid launch, which freed `event29`, and player 1's virtual pad was then created onto exactly that node. Player 3, whose controller had been `event29`, opened it and mirrored player 1's virtual pad instead. The device ended up literally named:

```
PD Pad 3 (PD Pad 1 (Xbox 360 Wireless Receiver))
```

One player's stick driving another player's game, and the real pad 3 driving nothing. If the rewrite creates a virtual device per player, this is reachable any time a controller reconnects during setup. What fixed it for me was resolving every pad to a stable id *before* creating any virtual device, then re-deriving the path from that id as each one is created.

**2. On a shared receiver there is no usable device identity except the USB interface.**

Four Xbox 360 pads on one wireless receiver report the same name, the same `045e:02a1`, the same `phys`, and an empty `uniq`. Every field you would normally identify a device by is identical across all four. The only thing that distinguishes them is the interface they hang off (`3-3:1.0`, `1.2`, `1.4`, `1.6`), which survives the pad power cycling. Worth knowing before designing a binding format, because "remember which controller this was" needs something to key on and the obvious fields are all useless here.

**3. Same path and same minor can still be a different device.**

Measured across one session, launch at 20:36:

```
20:35:55  event24 = input257   event29 = input259   event30 = input260
20:46:49  event29 -> input264
20:46:50  event30 -> input265
```

`event29` kept minor `13,93` throughout. The path did not change, the minor did not change, but the kernel input device was destroyed and rebuilt, so every open fd was dead. Anything caching an fd or trusting a minor number will silently stop receiving input rather than erroring.

**4. Virtual devices land above `event31`, and not everything scans that far.**

Once `event0` to `event31` are used up, uinput allocations jump to `event256`. Creating four virtual pads on a busy machine puts most of them up there. I tested SDL2 inside a replica of the sandbox with a device at `event256` and it enumerated fine both with `/run/udev` reachable and with it masked, so SDL is not the problem. Unity's Rewired is: it does its own numeric scan and does not look past 31, which is how a pad on `event256` ends up invisible to a game while working everywhere else. If rebinding hands games a virtual device, that decides whether some titles see it at all.

**5. Two small ones for the tmpfs work.**

`jsN` is a separate file from `eventN` for the same device, so under a tmpfs it has to be bound alongside or anything using the legacy joystick API finds nothing. And `hidraw` is outside `/dev/input` entirely, so it still needs masking rather than omitting. Both are in the patch here.

Happy to test anything against four pads on a wireless receiver across three monitors, which is where all of the above turned up and is a bit of an edge case to hit otherwise. I will have a look at the Matrix room.

wunnr:
Gonna get this merged now as we don't have a solid ETA for when the big update with the input system rewrite will be ready. I'm not a fan of huge comment blocks, the code in question is fairly self-evident IMO so I'll be simplifying the comments.

PR #213, opened 2026-08-22
Fix KWin placement across multiple monitors

west1nat0r:
## Problem

The KWin layout scripts identify every window whose resource class is `gamescope` or `gamescope-kbm`, then tile those windows on whichever output KWin currently reports for them.

That loses two pieces of information PartyDeck already has at launch time:

- which Gamescope process belongs to the current PartyDeck session;
- which monitor was selected for each instance.

On a Plasma Wayland multi-monitor setup, a new Gamescope window can initially appear on a different output from the one selected in PartyDeck. The script then groups and lays it out on that initial output instead of moving it to the configured one. It can also reposition unrelated Gamescope windows. Finally, one instance on an output receives borderless output-sized geometry but is not placed in KWin fullscreen mode.

This was reproduced with two instances assigned to separate 3440x1440 and 2560x1440 displays. The original script did not reliably preserve that assignment or enter actual fullscreen.

## Solution

Generate the KWin script after spawning the Gamescope processes, when their PIDs are known:

- map each spawned Gamescope PID to the monitor selected for its instance;
- serialize that mapping and the two-player orientation as JSON into one runtime KWin template;
- match KWin windows by PID instead of resource class;
- move each matched window to its configured output before calculating the layout;
- accept both exact connector names and SDL names with an appended model suffix;
- calculate player counts independently for each target output;
- set real KWin fullscreen when an output has exactly one PartyDeck instance;
- ignore unrelated Gamescope windows;
- reconcile `workspace.stackingOrder` immediately when the script starts, so windows created before script loading are not missed;
- terminate already-started children if a later spawn or KWin script startup fails.

Horizontal and vertical layouts share one template; only the two-player geometry is selected from the generated configuration. The generated script remains under PartyDeck's existing temporary data directory and is removed by the existing launch cleanup.

## Commit history

The change is intentionally one atomic commit because the Rust launch ordering, generated configuration, and KWin template form one runtime contract:

- `fix: target KWin windows by launch PID`

## Verification

Manual end-to-end validation was performed on the v0.8.6-based implementation from which this change was rebased:

- KDE Plasma 6.7.4 on Wayland;
- two Gamescope instances;
- one instance assigned to a 3440x1440 display and one to a 2560x1440 display;
- both windows appeared on their selected monitors in actual fullscreen;
- unrelated windows were not included.

After rebasing and consolidating the template on current `main`:

- `cargo test` — 2 passed;
- `cargo build --release` — passed;
- `node --check res/splitscreen_kwin.js` — passed;
- a mocked KWin workspace smoke check passed multi-monitor fullscreen, horizontal two-player split, vertical two-player split, and SDL output-name suffix matching;
- `git diff --check origin/main` — passed.

The release build emits the existing warnings for optional Goldberg/UMU artifacts that are not downloaded in the local checkout.

## AI assistance disclosure

Prepared with AI assistance using oh-my-pi, models `openai-codex/gpt-5.6-sol`, `openai-codex/gpt-5.6-terra` and `openai-codex/gpt-5.6-lua` . AI was used for log analysis, implementation drafting and refinement, test drafting, complexity review, and preparation of this PR description. The author directed the scope, supplied the reproducer, and performed the real-hardware dual-monitor validation. The final diff and verification results were reviewed before submission.


west1nat0r:
Hi, just wanted to add some context myself.

I had gpt-5.6 sol and the smaler 5.6 models terra and luna work on this issue because I wanted to play Elden Ring co-op on two separate screens on CachyOS. Unfortunately, that didn’t work out of the box, and PartyDeck/Gamescope kept spawning the game as split-screen on a single monitor.

I know ai written PRs mostly are a pain for open-source projects and their maintainers, since they often contain a lot of questionable code, fix one issue while introducing several others.

So if you don’t want to deal with this PR, please feel free to close it without merging. I’m not a Rust developer, but since I had already spent the tokens debugging and patching this, I figured I might as well contribute it. :)

davidawesome02-backup:
Thanks! Interest in any issues is much appreciated, and this is a known issue with our current model. Due to the heavy limitation with our current design, but me and Blahkaey are currently working on a new system around more direct EGUI (our render framework) and gamescope interactions, allowing us to render a single window with the content of multiple gamescopes inside, as well as place it on any given monitor, or just show in windowed mode without the use of the overly complex and DE specific kwin scripts.  HOPEFULLY we can get this out soon, as it is currently waiting on us to finish up some changes with new handlers, and fixing the launch system, then some clean up (should be merged to main before we finish cleanup fully). I am not going to close this yet as I will look more at it in a bit, but in general, our new approach should solve this issue while working in a more "correct" fashion. We are just taking a long time as we both had other things going on.

I had attempted for a while to get KWIN script modifications to work well, but before had issue with the limited debug tools as well as terrible documentation on them. I will look more at this later, but from a breif look, you got stuck in the same injecting into a script by modifying it before starting trap that makes it strange. I am also not sure if I had issues with getting pids on wayland / X11 windows, but I may be mistaken.

While I am not a fan of AI code being overly used, I think it can be helpful if you review and actually use it as a tool to help make very small changes, or use it to learn with (as in using it to generate example code, or find how to do something you construct on a larger level). I personally try to limit use of it, but I think in general as long as changes are properly reviewed and understood they can be helpful to offer as changes. We are interested in commits for partydeck as long as good faith effort was put into their creation, so posting this is appreciated, even if we pass it over for the other changes we are making.

If you do happen to have a good understanding of linux / concepts related to this project, even if not specifically in rust, we are looking for opinions on a few changes in the matrix chat and have some occasional updates posted there on the status when I remember to post what I am working on in the new stuff :P.

Sorry if this message is confusing, writing this again in a rush.

Thanks again!

Issue #216, opened 2026-09-11
[Feature request] Shim steam.exe process


CodeShellDev:
**Is your feature request related to a problem? Please describe.**
Whilst trying to get a game to not refuse to start I discovered that it had an unusual way of checking if it was in a real steam env. It checked for a running process with the filename of steam.exe.

**Describe the solution you'd like**
Technically already possible via .bat file instead of the game exe in the handler‘s exec field.
It would be great to have a toggle for enabling this built-in!

**Describe alternatives you've considered**
Batch file that launches both game and shim.

**Additional context**
This not important at all and is meant for when no other high priority issues are open.

CodeShellDev:
Could be done with #218

PR #218, opened 2026-09-12
Add prelaunch option for setting up handlers

CodeShellDev:
## Description
I have added a `prelaunch` option for handlers, that enables handlers to perform pre-launch operations,
for example to patch the `GameAssembly.dll` or maybe to run a background task as a sidecar.

The prelaunch command is run before all the other `new_cmds` commands from `launch_cmds()`,
and is supplied some important and useful environment variables like `GAMEDIR` or `HANDLEDIR`.

## What's Changed
- added new field in handler creator (`Prelaunch`)
- added new field in json handler (`prelaunch`)
- changed `print_launch_cmds()` to better separate instances and also prelaunch from regular launch command
  args are now no longer listed in this section to prevent clutter and duplication
- launch and prelaunch cmds are now announced with `Executing [PROGRAM] [ARGS]` (this was what I mean with duplication)
- user is notified if prelaunch is malformed on start
- users are asked to confirm if they would like to import .pd2 file if it has a `prelaunch` script

## Deps
- added a new direct dependency `shell-words` to parse `prelaunch` and split it into `program` and `args` part

I do know that rust can be somewhat difficult regarding memory allocation etc., not sure if I touched anything like that, but please do take a thorough look at that specifically.

I am open to changes and refactors!

Thanks in advance!,
CSD out 🐢


davidawesome02-backup:
Hey, I will take a look at this tomorrow, but for now we may not accept this directly. As you know, I am in the middle of wayyy too many changes to note here, and am waiting on Blahkaey to rewrite the handler system in general to allow for more modifications like this and the #216 change you had made. I will still look at this for the current version, but just wanted to let you know that we do have plans for a larger rewrite.

Also, don't worry about memory safety in most cases unless you are writing unsafe rust or doing obviously messy things, you wont cause a memory issue in rust, the whole idea of rust is that it makes those harder than other languages due to how typing and memory management works. 

Good to see you here, will update more in a bit. If I forget or get distracted, feel free to ping me again.

CodeShellDev:
> Hey, I will take a look at this tomorrow, but for now we may not accept this directly. As you know, I am in the middle of wayyy too many changes to note here, and am waiting on Blahkaey to rewrite the handler system in general to allow for more modifications like this and the #216 change you had made. I will still look at this for the current version, but just wanted to let you know that we do have plans for a larger rewrite.
> 
> Also, don't worry about memory safety in most cases unless you are writing unsafe rust or doing obviously messy things, you wont cause a memory issue in rust, the whole idea of rust is that it makes those harder than other languages due to how typing and memory management works.
> 
> Good to see you here, will update more in a bit. If I forget or get distracted, feel free to ping me again.

Oh I am sorry to hear that, I didn‘t want to cause any chaos, but was unsure if there actually is a feature freeze…

I mean I didn‘t really change much so this could probably be added fairly quickly especially with the new handler system instead.

Would be honoured to implement the same in the new upcoming handler system (🎉),
for now it serves me quite well so it has already done its job.

CodeShellDev:
Hey there, friendly ping - @davidawesome02-backup

CodeShellDev:
I would then rework the way this works with the new systems to also allow for overriding launch commands / behaviour!

CodeShellDev:
Closed due to upcoming changes.

